PR #11, opened 2024-02-05
Fix linter errors

DomAyre:
Run lint.sh with -f

### DIFF
--- a/src/endpoints/proto/gen/hpke_pb.ts
+++ b/src/endpoints/proto/gen/hpke_pb.ts
@@ -19,7 +19,14 @@
 /* eslint-disable */
 // @ts-nocheck
 
-import type { BinaryReadOptions, FieldList, JsonReadOptions, JsonValue, PartialMessage, PlainMessage } from "@bufbuild/protobuf";
+import type {
+  BinaryReadOptions,
+  FieldList,
+  JsonReadOptions,
+  JsonValue,
+  PartialMessage,
+  PlainMessage,
+} from "@bufbuild/protobuf";
 import { Message, proto3 } from "@bufbuild/protobuf";
 
 /**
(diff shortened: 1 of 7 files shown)


src/endpoints/proto/gen/hpke_pb.ts
```diff
@@ -19,7 +19,14 @@
 /* eslint-disable */
```
beejones, line 19: we shouldn't touch machine generated files.

src/endpoints/proto/gen/tink_pb.ts
```diff
@@ -21,7 +21,14 @@
 /* eslint-disable */
```
beejones, line 21: same here

src/endpoints/proto/gen/hpke_pb.ts
DomAyre: Should we then update lint.sh to ignore them?

DomAyre: Done in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/11/commits/c19596e2ee3530eb66d4de78742d2bc0daec1302

beejones: they already have ignore tags but clearly these tags are ignored :smile:. There is a way to configure lint. I can have a look to that after the JWT work or if you want to look into that, that's ok for me.

DomAyre: I had a look into it, looks like the ignore tags are for tools other than prettier, so I think the simplest solution is to ignore files in the `/gen` directory which is also conceptually neat. 

As of https://github.com/microsoft/azure-privacy-sandbox-kms/pull/11/commits/d4393c4df8e66c6e171f42b4edd2195c0b09db91 the linter passes without any changes to the generated files, so I think we're good to merge this if you're happy

src/endpoints/proto/gen/tink_pb.ts
DomAyre: Discussion above

beejones (approved):

PR #19, opened 2024-02-26
Fixing make demo to always run

beejones:
Make starting and stopping of sandbox consistent in makefile

### DIFF
--- a/test/e2e-test/src/index.ts
+++ b/test/e2e-test/src/index.ts
@@ -167,7 +167,6 @@ class Demo {
     );
 
     Demo.assertField(member.name, response, "x", notUndefinedString);
-    Demo.assertField(member.name, response, "x", undefinedString);
     Demo.assertField(
       member.name,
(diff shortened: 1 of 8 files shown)


test/e2e-test/src/index.ts
```diff
@@ -172,26 +179,90 @@ class Demo {
...
+    } else {
+      throw new Error(
+        `🛑 [TEST FAILURE]: Initial key response message failed`
+      );
+    }
+  
+  
     // Wait for receipt to be generated
-    await Demo.sleep(5000);
+    await Demo.sleep(3000);
```
takuro-sato, line 253: Maybe it's worth to remove these sleep now to avoid future pain (in a different PR)?
I think we can just retry getting wrapped key while getting 202.

takuro-sato (approved):
Once it's tested with the internal dev repo, I think it's ready to merge

PR #23, opened 2024-04-15
Add pre-commit and hook installation

DomAyre:


.secrets.baseline
```diff
@@ -0,0 +1,239 @@
...
+      "path": "detect_secrets.filters.heuristic.is_sequential_string"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_swagger_file"
+    },
+    {
+      "path": "detect_secrets.filters.heuristic.is_templated_secret"
+    }
+  ],
+  "results": {
```
DomAyre, line 110: This is where we want to review

.secrets.baseline
```diff
@@ -0,0 +1,239 @@
...
+    "src/endpoints/KeyWrapper.ts": [
+      {
+        "type": "Base64 High Entropy String",
+        "filename": "src/endpoints/KeyWrapper.ts",
+        "hashed_secret": "3383828c227d6d0ad9ea4ca67684fc6c75686207",
+        "is_verified": false,
+        "line_number": 54
```
takuro-sato, line 117: Just a comment for an example response. Encrypted tink key.
We could replace the text with something like "AOeDD+K9avWgJPATp...." to make it look safer.

.secrets.baseline
```diff
@@ -0,0 +1,239 @@
...
+      }
+    ],
+    "src/endpoints/TinkKey.ts": [
+      {
+        "type": "Base64 High Entropy String",
+        "filename": "src/endpoints/TinkKey.ts",
+        "hashed_secret": "3c0e588bd5ceb5de9d26f21393f39e239df5a5fa",
+        "is_verified": false,
+        "line_number": 21
```
takuro-sato, line 126: Should be safe. It's a public key in a comment

.devcontainer/Dockerfile.devcontainer
```diff
@@ -46,6 +45,9 @@ RUN cp tinkey /usr/bin/
 RUN cp tinkey_deploy.jar /usr/bin/
 RUN rm tinkey tinkey_deploy.jar tinkey.bat $TINKEY_VERSION.tar.gz
 
+RUN pip install -r requirements.txt
+RUN pre-commit install
```
takuro-sato, line 49: Does this automatically add the hook to `git commit` in KMS's devcontainer?

DomAyre: Yes, and anything else which uses the image built from this dockerfile

takuro-sato: Sounds good. How about adding it to the dev repo's devcontainer as well so that it's triggered to from that image/codespace?

.secrets.baseline
KenGordon: We could just change the actual secret to BLAH_BLAH_BASE64_SECRET_0102030405 etc.

.secrets.baseline
KenGordon: We could just change the actual secret to BLAH_BLAH_BASE64_SECRET_0102030405 etc. (ditto)

.secrets.baseline
beejones: fixed

.secrets.baseline
beejones: fixed

beejones (approved):

PR #25, opened 2024-04-23
Move devcontainer image build stages into scripts

DomAyre:
This allows us to have docker files where scripts are called separately to improve caching performance, but also have images which call everything in one layer

This is useful because tools (especially acipolicygen) will run much faster on images with fewer layers


DomAyre:
Additional testing by creating a codespace and running `make demo && npm run test`

kapilvgit (approved):

PR #26, opened 2024-04-30
Re-organise docker images

DomAyre:
We have two very similar devcontainers defined, and the one tested in CI is not the standard one used when creating a codespace

Delete the devcontainer defined under `.github/` and point the test at the one under `.devcontainer`

2024-07-26 Update:
Adding a new dockerfile which produces a small(er) image for running KMS in ACI or locally but also is quick to rebuild

Trade offs:
- Since KMS relies on sandbox.sh to run, we still need to use the dev image
- Since dummy IDP is bundled with KMS the image needs nodejs


KenGordon (approved):
Looks good to me. Can Ronny try it out to be sure?

beejones (commented):
@DomAyre I assume you have tried rebuilding the container in codespaces and run the CI in privacy sandbox, correct?

beejones (commented):
I see there is a merge conflict because my change went in and I also did improvements to the docker files. Can you look to the files in main and improve these with your PR?

DomAyre:
I've resolved the merge conflict.

Regarding testing, I've rebuilt the codespace successfully and started the kms, however no privacy-sandbox-dev tests should be affected by this change since it has it's own dockerfile for the KMS, so I haven't touched that.

PR #34, opened 2024-07-11
Update MS JWT proposal when AAD keys roll

beejones:
When AAD rolls keys we need to update the MS JWT proposal with the new certificates. The change make sure that governance/jwt/set_jwt_ms_validation_policy_proposal.json is updated. 
This proposal can be checked in but not necessary because it will always be generated.


Makefile
```diff
@@ -81,10 +81,11 @@ propose-jwt-demo-validation-policy: ## 🚀 Deploy the JWT validation policy
 
 # Propose a new idp
 propose-jwt-ms-validation-policy: ## 🚀 Propose the AAD as idp
+propose-jwt-ms-validation-policy: ## 🚀 Propose the AAD as idp. If AAD rolled keys, governance/jwt/set_jwt_ms_validation_policy_proposal.json will be updated and can be checked in
 	@echo -e "\e[34m$@\e[0m" || true
-	@./scripts/generate_jwt_proposal_payload.sh  --proposal-file "${KMS_WORKSPACE}"/proposals/set_jwt_ms_validation_policy_proposal.json
-	@CCF_PLATFORM=${CCF_PLATFORM} ./scripts/submit_proposal.sh --network-url "${KMS_URL}" --proposal-file "${KMS_WORKSPACE}"/proposals/set_jwt_ms_validation_policy_proposal.json --certificate_dir "${KEYS_DIR}" --member-count ${MEMBER_COUNT}
-
+	@./scripts/generate_jwt_proposal_payload.sh  --proposal-file "${PWD}"/governance/jwt/set_jwt_ms_validation_policy_proposal.json
```
takuro-sato, line 86: Using pwd could break the command when you are running make command with `-C` option from a different directory than usual. 

```bash
-C dir, --directory=dir
            Change  to  directory  dir before reading the makefiles or doing anything else.  If multiple -C options are specified, each is interpreted relative to the previ
ous one: -C / -C
            etc is equivalent to -C /etc.  This is typically used with recursive invocations of make.
```

It would be nicer if you can avoid it since you are making it production ready

governance/jwt/set_jwt_ms_validation_policy_proposal.json
```diff
@@ -4,7 +4,7 @@
...
...\nd3cuZGlnaWNlcnQuY29tMSAwHgYDVQQDExdEaWdpQ2VydCBHbG9iYWwgUm9vdCBD\nQTAeFw0wNjExMTAwMDAwMDBaFw0zMTExMTAwMDAwMDBaMGExCzAJBgNVBAYTAlVT\nMRUwEwYDVQQKEwxEaWdpQ2VydCBJbmMxGTAXBgNVBAsTEHd3dy5kaWdpY2VydC5j\nb20xIDAeBgNVBAMTF0RpZ2lDZXJ0IEdsb2JhbCBSb290IENBMIIBIjANBgkqhkiG\n9w0BAQEFAAOCAQ8AMIIBCgKCAQEA4jvhEXLeqKTTo1eqUKKPC3eQyaKl7hLOllsB\nCSDMAZOnTjC3U/dDxGkAV53ijSLdhwZAAIEJzs4bg7/fzTtxRuLWZscFs3YnFo97\nnh6Vfe63SKMI2tavegw5BmV/Sl0fvBf4q77uKNd0f3p4mVmFaG5cIzJLv07A6Fpt\n43C/dxC//AH2hdmoRBBYMql1GNXRor5H4idq9Joz+EkIYIvUX7Q6hL+hqkpMfT7P\nT19sdl6gSzeRntwi5m3OFBqOasv+zbMUZBfHWymeMr/y7vrTC0LUq7dBMtoM1O/4\ngdW7jVg/tRvoSSiicNoxBN33shbyTApOB6jtSj1etX+jkMOvJwIDAQABo2MwYTAO\nBgNVHQ8BAf8EBAMCAYYwDwYDVR0TAQH/BAUwAwEB/zAdBgNVHQ4EFgQUA95QNVbR\nTLtm8KPiGxvDl7I90VUwHwYDVR0jBBgwFoAUA95QNVbRTLtm8KPiGxvDl7I90VUw\nDQYJKoZIhvcNAQEFBQADggEBAMucN6pIExIK+t1EnE9SsPTfrgT1eXkIoyQY/Esr\nhMAtudXH/vTBH1jLuG2cenTnmCmrEbXjcKChzUyImZOMkXDiqw8cvpOp/2PV5Adg\n06O/nVsJ8dWO41P0jmP6P6fbtGbfYmbW0W5BjfIttep3Sp+dWOIrWcBAI+0tKIJF\nPnlUkiaY4IBIqDfv8NZ5YBberOgOzW6sRBc4L0na4UU+Krk2U886UAb3LujEV0ls\nYSEY1QSteDwsOoBrp+uvFRTp2InBuThs4pFsiv9kuXclVzDAGySj4dzp30d8tbQk\nCAUw7C29C79Fv1C5qfPrmAESrciIxpg0X40KPMbp1ZWVbd4=\n-----END CERTIFICATE-----"
```
takuro-sato, line 7: Looks like we just got another hard-coded cert which will expire in future again. What will happen when they do key roll?
I feel like I'm missing something...

beejones: This will automatically be updated during each start of kms when we run setup.  If it changes (keys rolled) we should check it in.

PR #35, opened 2024-07-11
Remove ext_expires_in to match managed identity JWT

takuro-sato:
- Remove ext_expires_in to match managed identity JWT
- Change type of expires_in to match managed identity JWT (The original is correct in terms of RFC as Ronny suggests in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/35#discussion_r1675422972)

Related PR: https://github.com/microsoft/privacy-sandbox-dev/pull/167

### DIFF
--- a/governance/jwt/set_jwt_ms_validation_policy_proposal.json
+++ b/governance/jwt/set_jwt_ms_validation_policy_proposal.json
@@ -27,11 +27,11 @@
           "appidacr": "2",
           "idp": "https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/",
           "idtyp": "app",
-          "oid": "da96e2ae-76bc-402e-abfe-2c3bd3bdf226",
-          "sub": "da96e2ae-76bc-402e-abfe-2c3bd3bdf226",
+          "oid": "049a48cd-7652-4159-82cf-7a2a42248b38",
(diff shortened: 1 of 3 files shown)


test/utils/jwt/src/index.ts
beejones, line 1: https://www.rfc-editor.org/rfc/rfc6749. This is the oauth2 standard and it defines a number for expires_in.

governance/jwt/set_jwt_ms_validation_policy_proposal.json
beejones, line 1: It is great to use the values for CI in our default policy but we might need to support multiple apps with one KMS, hence we need to support multiple values for these fields. I will make that change in KMS.

beejones (approved):

test/utils/jwt/src/index.ts
takuro-sato: Thanks, letting me know. That's interesting.

>  we could possibly support a string and number in parsing

That sounds like a good idea. I'll try to see if I can do it.

takuro-sato: Handled in a PR for data-plane repo here https://github.com/KenGordon/data-plane-shared-libraries/pull/33/commits/b69c3db531301fef8014e4b03182ca4d27bfe301

PR #38, opened 2024-07-19
Beejones/updating npm packages

beejones:
Update package.json packages
Improve container


src/endpoints/keyEndpoint.ts
```diff
@@ -15,6 +15,9 @@ import { hpkeKeyIdMap, hpkeKeysMap } from "../repositories/Maps";
 import { ServiceRequest } from "../utils/ServiceRequest";
 import { Logger } from "../utils/Logger";
 
+// Enable the endpoint
```
takuro-sato, line 18: Why does it need to be moved here?

tsconfig.json
```diff
@@ -11,6 +11,7 @@
     "sourceMap": false,
     "resolveJsonModule": true,
     "types": ["@types/node"],
+    "strict": false,
```
takuro-sato, line 14: Why can't we use strict:true ?

takuro-sato (commented):
Can you put more context please? There are lot of changes that the reason is unclear to me. 
I put some example questions.

tsconfig.json
beejones: I am currently running the CI with strict: true. I did this experiment because of failing CI's in ACI. After this last run we can merge this version and we can remove the install of rollup from privacy-sandbox-dev

src/repositories/LastestItemStore.ts
```diff
@@ -45,7 +45,7 @@ export class LastestItemStore<K extends number, T> {
   }
 
   public receipt(id: K) {
-    const version = this.store.getVersionOfPreviousWrite(id);
+    const version = this.store.getVersionOfPreviousWrite(id) || 0;
```
takuro-sato, line 48: Is it really OK to set version=0?
I feel like we should return "not found" instead, but not sure

src/authorization/AuthenticationService.ts
```diff
@@ -47,8 +47,11 @@ export class AuthenticationService implements IAuthenticationService {
    */
   public isAuthenticated(
     request: ccfapp.Request<any>,
-  ): [ccfapp.AuthnIdentityCommon, ServiceResult<string>] {
-    let caller: ccfapp.AuthnIdentityCommon;
+  ): [
+    ccfapp.AuthnIdentityCommon | undefined,
+    ServiceResult<string> | undefined,
```
takuro-sato, line 52: Shouldn't we always return `ServiceResult<string>`? I think we want `validator.validate(request)` without `?` after confirming `validator` is not null/undefined.

src/endpoints/KeyWrapper.ts
```diff
@@ -212,7 +217,7 @@ export class KeyWrapper {
   private static getEncryptedKeyMaterial(
     wrappingKey: ArrayBuffer | undefined,
     payload: IKeyItem,
-  ): [string, string] {
+  ): [string, string | undefined] {
     Logger.debug(`getEncryptedKeyMaterial: `, payload);
     const receipt = payload.receipt;
     delete payload.receipt;
```
takuro-sato, line 223: What is the condition that `payload.receipt` is undefined? I'm wondering if we should allow undefined for the response?

src/authorization/jwt/MsJwtProvider.ts
```diff
@@ -18,7 +18,7 @@ export const authorizeJwt = (
   issuer: string,
   identity: ccfapp.JwtAuthnIdentity,
 ): ServiceResult<string> => {
-  const policy = JwtValidationPolicyMap.read(issuer);
+  const policy = JwtValidationPolicyMap.read(issuer) || {};
```
takuro-sato, line 21: Shouldn't we return "issuer is not registered" error rather than using empty policy?
I guess I'm missing something, but I feel like there would be no check if issuer doesn't exist in JwtValidationPolicyMap?

.github/devcontainer.json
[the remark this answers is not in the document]
takuro-sato: I see. I just didn't know about `devcontainers/ci@v0.3`. 

> Are you saying we can have one for both?

I guess it's preferable, but what's your view on that? At least I don't think we need to address it within this PR.

package.json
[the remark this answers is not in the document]
beejones: I added an issue: https://github.com/microsoft/azure-privacy-sandbox-kms/issues/39
Looks like a great suggestion to add.

src/attestation/SnpAttestationClaims.ts
```diff
@@ -83,7 +83,7 @@ export class SnpAttestationClaims {
     // report_id_ma
     // reported_tcb
     // chip_id
-    // committed_tcb
+    // committed_tcbx`
```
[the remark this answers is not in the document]
beejones: no idea how this happened

scripts/set_python_env.sh
```diff
@@ -11,6 +11,7 @@ if [ ! -d "$PYTHON_VENV" ]; then
 	python3.8 -m venv $PYTHON_VENV
 	echo "Activating virtual environment and installing dependencies..."
 	source $PYTHON_VENV/bin/activate
+	pip install wheel
 	pip install -U -r ./requirements.txt
```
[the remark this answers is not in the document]
beejones: good question. I used to only have it in requirements and it generated an error. This solves the error.

src/repositories/LastestItemStore.ts
beejones: Added a test for undefined

src/authorization/AuthenticationService.ts
beejones: I added a failed result so undefined is never returned.

src/endpoints/KeyWrapper.ts
beejones: if receipt is undefined, the delete operation has no effect.

scripts/set_python_env.sh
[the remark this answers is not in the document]
takuro-sato: Why does it solve the problem?

src/authorization/jwt/MsJwtProvider.ts
beejones: good suggestion

src/endpoints/KeyWrapper.ts
takuro-sato: Thanks for letting me know that, but my question is really:
Is it possible getEncryptedKeyMaterial is called with `payload.receipt === undefined`? If so when is it?

I'm asking this because return type `[string, string]` is better than `[string, string | undefined]` if possible. Caller wouldn't have to handle undefined.

takuro-sato: I found the answer:

In key(), you can find:

```
// Get receipt if available
  if (receipt !== undefined) {
    keyItem.receipt = receipt;
    Logger.debug(`Key->Receipt: ${receipt}`);
  } else {
    return ServiceResult.Accepted();
  }
```.

So both `payload.receipt === undefined` and `payload.receipt !== undefined`, is expected.

scripts/set_python_env.sh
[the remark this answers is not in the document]
beejones: Interesting. After fixing the versions of the packages, the error is gone and there is no more need to install wheel first. This must have been a conflict between installed packages

takuro-sato:
Why do we need [2fe6bdf](https://github.com/microsoft/azure-privacy-sandbox-kms/pull/38/commits/2fe6bdfbd0727a7a37cd77b920780764bf2a7166) within this PR?

beejones:
> Why do we need [2fe6bdf](https://github.com/microsoft/azure-privacy-sandbox-kms/pull/38/commits/2fe6bdfbd0727a7a37cd77b920780764bf2a7166) within this PR?

I believed we needed it because I saw JWT failures in the container log. That's why I added it. We need  this change anyway so lets keep it

takuro-sato (approved):
We have a TODO https://github.com/microsoft/azure-privacy-sandbox-kms/issues/40, but the code looks good to me.

PR #51, opened 2024-08-14
Add description in hearthbeat, cleanup mCCF setup

beejones:
We show the Settings description in hearthbeat to show the version of KMS.
Small cleanup of mCCF setup

### DIFF
--- a/src/endpoints/kms.ts
+++ b/src/endpoints/kms.ts
@@ -5,9 +5,15 @@ import * as ccfapp from "@microsoft/ccf-app";
 import { ServiceResult } from "../utils/ServiceResult";
 import { enableEndpoint } from "../utils/Tooling";
 import { ServiceRequest } from "../utils/ServiceRequest";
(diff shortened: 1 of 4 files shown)


src/endpoints/kms.ts
```diff
@@ -17,13 +23,24 @@ enableEndpoint();
...
 
   // check if caller has a valid identity
   const [policy, isValidIdentity] = serviceRequest.isAuthenticated();
   if (isValidIdentity.failure) return isValidIdentity;
 
-  return ServiceResult.Succeeded<ccfapp.AuthnIdentityCommon>(policy!);
+  const settings = Settings.loadSettings();
+  const description = [
+    settings.settings.service.name,
+    settings.settings.service.description,
```
takuro-sato, line 37: Maybe we want to put that `description` is a public information in README so that users don't accidentally put things that shouldn't be public?

takuro-sato (approved):

PR #126, opened 2024-09-18
Add script to add hostdata value to the key release policy

beejones:
The new scripts/add_hostdata_keyreleasepolicy.sh allow us to add hostdata to  the key release policy.

Usage:
scripts/add_hostdata_keyreleasepolicy.sh --network_url $KMS_URL --certificate_dir $KEYS_DIR --hostdata 73973b78d70cc68353426de188db5dfc57e5b766e399935fb73a61127ea26d20

This adds the hostdata in our test attestation to the key release policy

### DIFF
--- a/scripts/add_hostdata_keyreleasepolicy.sh
+++ b/scripts/add_hostdata_keyreleasepolicy.sh
@@ -1,0 +1,98 @@
+#!/bin/bash
+
+# Copyright (c) Microsoft Corporation.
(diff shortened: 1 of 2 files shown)


takuro-sato (approved):
From chat with Ronny:
- To have multiple values for hostdata, we just can run the script multiple time with different value.
- The script is tested with mccf

PR #137, opened 2024-09-24
Add key refresh test for  mCCF KMS

takuro-sato:
Run a CI to exercise basic KMS operation to generate and fetch keys.
The new script was tested locally as well.

Testing: https://github.com/microsoft/privacy-sandbox-dev/actions/runs/11015588720


scripts/setup_local.sh
```diff
@@ -3,7 +3,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-set -euo pipefail
```
takuro-sato, line 6: If it's designed to be  used through `source` command (or `.`), I think it is better to remove it.
If some command fails after running `. scripts/setup_local`, the shell crashes

takuro-sato:
New test run: https://github.com/microsoft/privacy-sandbox-dev/actions/workflows/ci.yml

beejones (approved):

PR #154, opened 2024-10-08
Fixing deployment bugs and adding additional docs

beejones:
We tested a new deployment and this suggested some small changes to the Makefile and docs.


beejones:
Testing:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/11236182842

docs/README.md
```diff
@@ -0,0 +1,15 @@
+# Assumptions
```
takuro-sato, line 1: Maybe we might want to add a comment about creating an AKV instance here just to  make it a bit easier to understand.

At a glance I was not sure `scripts/create_member_cert_akv.sh` creates an AKV instance or I need to make one. You can figure it out just by trying, but it would nice if it's clarified

takuro-sato (approved):

docs/README.md
beejones: done

PR #155, opened 2024-10-09
Fix for AKV signing and update docs

beejones:
The PR fixes the submission of proposals signed with an AKV member key. We were missing the voting on proposals.
Small edits to docs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,15 +109,16 @@ refresh-key: ## 🚀 Refresh a key on the instance
 	$(call check_defined, KMS_URL)
 	@CCF_PLATFORM=${CCF_PLATFORM} sleep 20;curl "${KMS_URL}"/app/refresh -X POST --cacert "${KEYS_DIR}"/service_cert.pem  -H "Content-Type: application/json" -i  -w '\n'
 
-set-constitution: start-host-idp ## Set new custom constitution
+set-constitution: ## Set new custom constitution
 	@echo -e "\e[34m$@\e[0m" || true
 	$(call check_defined, KMS_URL)
 	$(call check_defined, KEYS_DIR)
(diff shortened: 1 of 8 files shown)


beejones:
Testing:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/11256127843

takuro-sato:
How can KMS members vote to a proposal that were created by another member? Is it future work or already covered by the current code?

Makefile
```diff
@@ -109,15 +109,16 @@ refresh-key: ## 🚀 Refresh a key on the instance
...
-set-constitution: start-host-idp ## Set new custom constitution
+set-constitution: ## Set new custom constitution
 	@echo -e "\e[34m$@\e[0m" || true
 	$(call check_defined, KMS_URL)
 	$(call check_defined, KEYS_DIR)
-	 if ! cmp -s ${KMS_WORKSPACE}/sandbox_common ${KEYS_DIR}; then \
-	 	@sleep 5
+	if [ "${KMS_WORKSPACE}/sandbox_common" != "${KEYS_DIR}" ]; then \
+		echo "Copying files for constitution"; \
+		@sleep 5; \
```
takuro-sato, line 118: That's great you clarified the purpose.
Can we avoid using sleep or make an issue for that? I faced 'missing constitution' errors a lot in the past, maybe this is the reason.

beejones: correct. We needed to copy these files manually. Now it will be automatically

beejones: There us a comment in the AKV signing script that only one member signs the proposal. We need an additional script only for the voting. I created an issue:
https://github.com/microsoft/azure-privacy-sandbox-kms/issues/156

takuro-sato (approved):

PR #166, opened 2024-10-17
Add system level testing which can be run locally and in C-ACI

DomAyre:
To be down-streamed to forks.

- Adds a docker-compose service for the CCF sandbox which can be run locally and will be runnable in ACI. This separation will make subbing other CCF deployments (local/mccf/azure-cleanroom/etc.) trivial
- Adds makefile targets to bring the service up and down and then KMS code/constitution etc. can be applied.
- Tests use these to spin up KMS's then tests endpoints

Todo:
- [x] Add more scenarios for all endpoints
- [x] Implement setting up KMS constitution including test JWT issuer
- [x] Implement running in ACI

UPDATE: deferring for future PR

- [ ] Handle flakiness in deploying KMS to C-ACI CCF network
- [ ] Add a cleanup workflow for C-ACI deployments
- [ ] Add tests for good and bad attestations
- [ ] Add tests for live attestations 
- [ ] Add tests for unwrapKey

Issues found so far (currently more nitpicks and probably aren't blockers):
- [x] #168 
- [ ] #167 
- [ ] #175 
- [ ] #176 
- [ ] #177
- [ ] #179


DomAyre:
Since running in ACI for CI is a bit flaky and we don't want to block fixing the issues, I suggest we merge this as is @beejones  if you're happy with this?

beejones (approved):

PR #198, opened 2024-11-13
Tail kms logs instead of /dev/null

DomAyre:
This change adds logs from the CCF network and KMS to the container logs.

This gives us a better understanding of what the KMS is doing, previously we were only emitting logs from the setup process


KenGordon:
For context. There will be a PR from Ronny improving the logging. In order to check that it doesn't do anything unexpected we need to be able to view the logs generated by the CI. This PR allows for that.

PR #202, opened 2024-11-15
print log and add daily schedule for CI

beejones:
After the e2e tests we print out the CCF log file.
Add a daily schedule to run CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,9 @@ jobs:
           runCmd: |
              scripts/set_python_env.sh && pip install -r requirements.txt
              make demo
-             cat workspace/sandbox_0/out
+
+      - name: Print CCF logs
+        run: cat workspace/sandbox_0/out
 
   system:
     secrets: inherit # pragma: allowlist secret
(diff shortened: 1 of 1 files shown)


beejones:
tests:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/11857560675

DomAyre:
I initially assumed that e2e tests would run multiple KMS instances (one per test) so I was concerned that printing the out file would only show the last KMS instances logs.

However it looks like we run one KMS instance for all e2e tests? If so, having tests which don't start with a clean slate could result in brittle tests since they might rely on a previous test doing something specific.

However that isn't an issue with this PR so happy to merge this

beejones:
> I initially assumed that e2e tests would run multiple KMS instances (one per test) so I was concerned that printing the out file would only show the last KMS instances logs.
> 
> However it looks like we run one KMS instance for all e2e tests? If so, having tests which don't start with a clean slate could result in brittle tests since they might rely on a previous test doing something specific.
> 
> However that isn't an issue with this PR so happy to merge this

this is intentional because kms goes in different states. E.g. no keys -> initial keys 202 -> receipt generated

PR #215, opened 2024-11-21
Add support for Azure Cleanroom based CCF

DomAyre:
> TEMP: Targets the `az-cleanroom` branch until that's merged

### Motivation
This allows us to test against a production-like mCCF replacement, and will allow us to validate it against a realistic use case for us

### Changes
- [x] Add an `scripts/az-cleanroom-aci` dir which installs the extension and brings up the network
- [x] Provide a deployment name to `conftest.py` for deploying the network
- [x] Handle up script having logs other than the resulting env
- [ ] Implement a way to remove all resources deployed by `up.sh` (maybe `az cleanroom ccf network down`?)
- [ ] Resolve trusting the service cert (temp using -k)
- [ ] Add runs of this to CI


.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,64 @@
+name: Cleanup
+
+permissions:
+  id-token: write
+  contents: read
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
```
beejones, line 10: Doing this every night is problematic. Could we do the cleanup on Sunday?
Scenario:
We have large builds/tests. Start a full build/test before shutting down the day to inspect the next morning.

Or maybe we can do cleanup for containers older than e.g. 2 days

.github/workflows/endpoint-test.yml
```diff
@@ -41,5 +45,9 @@ jobs:
...
       - name: Run System Tests
         env:
-            TEST_ENVIRONMENT: ${{ matrix.env }}
-        run: pytest -sv test/system-test/test_${{ inputs.endpoint }}.py
+          TEST_ENVIRONMENT: ccf/${{ matrix.config.env }}
+        run: |
+          source ./scripts/jwt_issuer/up.sh
+          pytest -sv \
+            test/system-test/test_${{ inputs.endpoint }}.py \
+            -n ${{ matrix.config.threads }}
```
beejones, line 53: cool. Can we use -n on our other CI's?

services/cacitesting.env
```diff
@@ -1,5 +1,5 @@
 # Core Environment Variables
-export SUBSCRIPTION=85c61f94-8912-4e82-900e-6ab44de9bdf8
+export SUBSCRIPTION=7ca35580-fc67-469c-91a7-68b38569ca6e
```
beejones, line 2: Shouldn't this be a secret?

.github/workflows/endpoint-test.yml
DomAyre: If by other CI's you mean using the `sandbox_local` environment, no because it's currently got hardcoded endpoint ports so if you tried to run multiple concurrently they would clash.

services/cacitesting.env
DomAyre: Why? It's just an ID you can't do anything with it without sufficient role based access

.github/workflows/cleanup.yml
beejones: This is a plead for not doing nightly cleanups for any of our repos. At least wait one day so if you come back in the morning you can still inspect the failing tests

services/cacitesting.env
beejones: OK fair

beejones (approved):

.github/workflows/cleanup.yml
DomAyre: The cleanup is before the nightly testing, so you can inspect the resources deployed by a given failing run for a full work day before it's cleaned up. And of course the logs in actions itself should have a reasonable amount of information to be useful too and that lasts much longer

PR #224, opened 2024-11-28
Run `az-cleanroom-aci` tests in parallel

DomAyre:
This is currently blocked on https://github.com/Azure/azure-cleanroom/pull/12 so raising as a separate PR


beejones (commented):
My only thought is on the naming az-cleanroom-aci. Aren't we using a project name for something which will be generic?

PR #232, opened 2024-12-03
update testing README

beejones:
Explain the testing concept in the README

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -1,6 +1,14 @@
 # Testing Strategy
 
-## Overview
+
+## Table of Contents
+
+- [Introduction](#introduction)
+- [Unit Testing](#unit-testing)
+- [End to End Testing (System Tests)](#end-to-end-testing-system-tests)
+- [Quick End to End Testing](#quick-end-to-end-testing)
+
+## Introduction
 
 The purpose of this document is to describe how the application is being tested to assure quality and coverage.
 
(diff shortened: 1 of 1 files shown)


test/README.md
```diff
@@ -1,93 +1,62 @@
 # Testing Strategy
 
-## Overview
+
+## Table of Contents
+
+- [Introduction](#introduction)
+- [Unit Testing](#unit-testing)
+- [End to End Testing (System Tests)](#end-to-end-testing-system-tests)
```
DomAyre, line 8: I think this naming scheme could be clearer, why not just call them what they're called in the code? System Tests and E2E Tests

test/README.md
```diff
@@ -1,93 +1,62 @@
...
+- [Quick End to End Testing](#quick-end-to-end-testing)
+
+## Introduction
 
 The purpose of this document is to describe how the application is being tested to assure quality and coverage.
 
 The application testing strategy depends on two main types of testing to achieve the objective:
 
 - Unit testing
 - End to end (e2e) testing
```
DomAyre, line 18: This is especially confusing, you've called it e2e which is that the tests you refer to as quick e2e tests are called in the code

test/README.md
DomAyre: Also quick e2e might be confusing since you can easily run a single heartbeat system test which would be faster than running the e2e tests. I appreciate currently the e2e tests are faster if you run them all but it isn't a fundamental feature of the style of tests.

If we run CCF networks locally in parallel which we might add then system tests would be faster than e2e tests, so the name could also end up being misleading

DomAyre (approved):

PR #233, opened 2024-12-03
Add AKV support to system tests

DomAyre:
### Motivation

In order to test azure-cleanroom based CCF networks, we want to have testing with as representative systems as possible. Production systems will store member identities in something like AKV rather than locally, so we want test coverage of this

### Changes 
- [x] Add lifecycle scripts to bring AKVs up and down 211127541e24d9c39db1919fe82a80ded415a72e
- [x] Update propose.sh to allow using AKV keys bd9e3e1d1ca2bcaf9e0b64e0d1d3bbf2ba13281f
- [x] Add ability to run tests using an AKV key 6b866ddb9aa843987bcb4ad14c6708736ae75fb4
- [x] Add tests using AKV keys 9f4ea4cdbad67dd501fae2e039d99a5830bcfb75
- [x] Update cleanup job to cleanup AKVs that failed to be deleted during runs 293ac8f5de1ea1f33d1699618ca48c647f5525d3

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -66,3 +66,31 @@ jobs:
               --resource-group "$RESOURCE_GROUP" \
               --yes
           done
+
+  cleanup-akv:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Log into Azure
+        uses: azure/login@v2
+        with:
+          client-id: ${{ secrets.MANAGED_ID_CLIENT_ID }}
+          tenant-id: ${{ secrets.MANAGED_ID_TENANT_ID }}
+          subscription-id: 7ca35580-fc67-469c-91a7-68b38569ca6e
+
+      - run: |
+          source services/cacitesting.env
+          key_vaults=$(az keyvault list \
+            --resource-group "$RESOURCE_GROUP" \
+            --query "[].name" \
+            -o tsv \
+          )
+
+          for kv in $key_vaults; do
+            echo "Deleting Key Vault: $kv"
+            az keyvault delete \
+              --name "$kv" \
+              --resource-group "$RESOURCE_GROUP" > /dev/null 2>&1
+          done
--- a/scripts/akv/down.sh
+++ b/scripts/akv/down.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+akv-down() {
+    set -e
+
+    REPO_ROOT="$(realpath "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../..")"
+
+    source $REPO_ROOT/services/cacitesting.env
+    AKV_NAME=${AKV_NAME:-$1}
+    if [ -z "$AKV_NAME" ]; then
+        read -p "Enter AKV name: " AKV_NAME
+    fi
+
+    az keyvault delete \
+        --resource-group $RESOURCE_GROUP \
+        --name $AKV_NAME
+
+    unset AKV_NAME
+
+    set +e
+}
+
+akv-down "$@"
--- a/scripts/akv/down.sh
+++ b/scripts/akv/down.sh
@@ -9,16 +9,16 @@ akv-down() {
     REPO_ROOT="$(realpath "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../..")"
 
     source $REPO_ROOT/services/cacitesting.env
-    AKV_NAME=${AKV_NAME:-$1}
-    if [ -z "$AKV_NAME" ]; then
-        read -p "Enter AKV name: " AKV_NAME
+    AKV_VAULT_NAME=${AKV_VAULT_NAME:-$1}
+    if [ -z "$AKV_VAULT_NAME" ]; then
+        read -p "Enter AKV name: " AKV_VAULT_NAME
     fi
 
     az keyvault delete \
         --resource-group $RESOURCE_GROUP \
-        --name $AKV_NAME
+        --name $AKV_VAULT_NAME
 
-    unset AKV_NAME
+    unset AKV_VAULT_NAME
 
     set +e
 }
--- a/scripts/akv/key-import.sh
+++ b/scripts/akv/key-import.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+akv-key-import() {
+    set -e
+
+    AKV_NAME=${AKV_NAME:-$1}
+    if [ -z "$AKV_NAME" ]; then
+        read -p "Enter AKV name: " AKV_NAME
+    fi
+    export AKV_NAME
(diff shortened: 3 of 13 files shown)


.github/workflows/cleanup.yml
```diff
@@ -65,4 +65,32 @@ jobs:
...
+  cleanup-akv:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Log into Azure
+        uses: azure/login@v2
+        with:
+          client-id: ${{ secrets.MANAGED_ID_CLIENT_ID }}
```
beejones, line 79: let's use the best practice to document how to get these values. In this case you can refer to the doc in docs describing these.

.github/workflows/cleanup.yml
```diff
@@ -65,4 +65,32 @@ jobs:
...
+          tenant-id: ${{ secrets.MANAGED_ID_TENANT_ID }}
+          subscription-id: 7ca35580-fc67-469c-91a7-68b38569ca6e
```
beejones, line 81: we already have a var for subscription.

.github/workflows/cleanup.yml
```diff
@@ -65,4 +65,32 @@ jobs:
...
+
+      - run: |
+          source services/cacitesting.env
+          key_vaults=$(az keyvault list \
+            --resource-group "$RESOURCE_GROUP" \
```
beejones, line 86: we should use a specific KV name for our tests to avoid deleting of existing keys

.github/workflows/cleanup.yml
DomAyre: It has a different value since we need to temporarily use `Parakeet-Services-NonProd` which allows us to deploy storage accounts with shared key access until our exception is granted

.github/workflows/cleanup.yml
DomAyre: In the resource group in question, we only deploy AKV instances for tests, so it's fine to delete everything. This exact logic applies to the container instances and storage accounts in the jobs above this one

scripts/akv/key-import.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+akv-key-import() {
+    set -e
+
+    AKV_NAME=${AKV_NAME:-$1}
```
beejones, line 9: AKV_VAULT_NAME

scripts/akv/down.sh
```diff
@@ -0,0 +1,26 @@
...
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+akv-down() {
+    set -e
+
+    REPO_ROOT="$(realpath "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../..")"
+
+    source $REPO_ROOT/services/cacitesting.env
+    AKV_NAME=${AKV_NAME:-$1}
```
beejones, line 12: AKV_VAULT_NAME for clarity.

.github/workflows/cleanup.yml
DomAyre: Can this be a separate PR? There's several places which uses this to login some of which are already on main. 

I'm also pro having the code self document to avoid desyncing between docs and code, and I don't know how better I can describe where to get the client id than `MANAGED_ID_CLIENT_ID` other than just converting that name into prose e.g. "_Go to the managed identity and use it's client id_"

scripts/akv/key-import.sh
```diff
@@ -0,0 +1,33 @@
...
+    fi
+    export AKV_NAME
+
+    AKV_KEY_NAME=${AKV_KEY_NAME:-$1}
+    if [ -z "$AKV_KEY_NAME" ]; then
+        read -p "Enter AKV Key name: " AKV_KEY_NAME
+    fi
+    export AKV_KEY_NAME
+
+    az keyvault key import \
```
beejones, line 21: why do we need to import the private key? The best practice is that the private keys is not exportable.
The signature will be generated on AKV.

scripts/ccf/propose.sh
```diff
@@ -3,23 +3,69 @@
...
 
+sign_proposal() {
+
+    proposal=$1
+
+    if [[ $USE_AKV == false ]]; then
+        ccf_cose_sign1 \
+            --content $proposal \
+            --signing-cert ${KMS_MEMBER_CERT_PATH} \
+            --signing-key ${KMS_MEMBER_PRIVK_PATH} \
```
beejones, line 14: you should use ccf_cose_sign1_prepare/finish to sign on AKV.
See sign_proposal_akv.sh

beejones (commented):
It looks like you are using AKV as a key store while we want to do the signature on AKV.

scripts/akv/key-import.sh
DomAyre: If we generate credentials on the AKV itself we need a separate member in the CCF network as credentials for the initial member is generated locally.

While the use case of a new member using credentials which never leave AKV is a great use case, it's a future step.

scripts/ccf/propose.sh
DomAyre: This code is if `USE_AKV == false` please see the else case three lines down

scripts/akv/key-import.sh
DomAyre: Sure I'll do this

scripts/akv/key-import.sh
DomAyre: https://github.com/microsoft/azure-privacy-sandbox-kms/issues/235

scripts/akv/key-import.sh
DomAyre: Actually, doesn't the V in AKV already stand for vault?

.github/workflows/cleanup.yml
beejones: so cleanup will only be scheduled?

scripts/akv/key-import.sh
beejones: There a many AKV artificate AKV_CERTIFICATE_NAME, AKV_KEY_NAME

So it is clearer to state we declare the vault name in same way we do this for the certificate.

DomAyre: AKV_NAME stands for Azure Key Vault Name
AKV_KEY_NAME stands for Azure Key Vault Key Name

Are you sure you want AKV_VAULT_NAME which stands for Azure Key Vault Vault Name? I think it's completely redundant, but it's not a big deal either way

beejones (approved):

scripts/akv/key-import.sh
DomAyre: https://github.com/microsoft/azure-privacy-sandbox-kms/pull/233/commits/fa3f24570c49c3fb9538db1c25a5f5aa39658ae0

scripts/akv/down.sh
DomAyre: https://github.com/microsoft/azure-privacy-sandbox-kms/pull/233/commits/fa3f24570c49c3fb9538db1c25a5f5aa39658ae0

PR #236, opened 2024-12-05
Document ci vars and secrets for endpoint-test

beejones:
Document how to set the necessary secrets and vars in CI's

### DIFF
--- a/.github/workflows/endpoint-test.yml
+++ b/.github/workflows/endpoint-test.yml
@@ -32,8 +32,8 @@ jobs:
         with:
           # Use a managed identity to authenticate to Azure
           # Use properties such as client-id, tenant-id, and subscription-id in the secrets and vars
-          # The managed identity should have a federated credential with subject identiier repo:<organisation>/<repo>:pull_request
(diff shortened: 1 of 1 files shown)


.github/workflows/endpoint-test.yml
```diff
@@ -30,6 +30,10 @@ jobs:
       - name: Log into Azure
         uses: azure/login@v2
         with:
+          # Use a managed identity to authenticate to Azure
+          # Use properties such as client-id, tenant-id, and subscription-id in the secrets and vars
+          # The managed identity should have a federated credential with subject identiier repo:<organisation>/<repo>:pull_request
+          # Use subject identiier repo:<organisation>/<repo>:ref:refs/heads/<branch> for running manual CI's
```
DomAyre, line 36: I don't think the word repo makes sense here? It's just a subject identifier

beejones: for creating the federated credential you need to enter the organization and repo to produce the subject identifier. I believe it is good to clarify that.

DomAyre: Oh sorry you're right the line wrapping on my view made it look like repo wasn't part of the identifier but of course it is

DomAyre (approved):

PR #240, opened 2024-12-05
Add scripts for adding members to a CCF network

DomAyre:
### Motivation
This is part of our proving out of different CCF network methods, namely azure-cleanroom. `az cleanroom ccf network up` creates a simple starter network with only one member, and production networks will have multiple members, so we need to be able to add members after network creation

### Changes

- [x] Refactor signing CCF proposals into standalone function d7b983926d56418fbe5cf792b10f198e8e0171da
- [x] Add scripts to add new members to the network ec32e22d58a063fe6aa3e242de5b2bb04ba31866
- [x] Add script to switch which member you operate with de9c770f9587b7f1c80734a3d99dd67d4fef8486
- [x] Add testing
  - [x] Add testing where existing members vote on new member joining (required #242)
- [ ] (If az cleanroom support delivered in time) Use az cleanroom functionality instead of custom scripts


scripts/ccf/member/add.sh
```diff
@@ -0,0 +1,87 @@
+#!/bin/bash
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+REPO_ROOT="$(realpath "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../../..")"
+
+ccf-member-add() {
+    set -e
```
beejones, line 9: should we delete create_member_cert_akv.sh?

DomAyre: Are there any existing workflows that anyone has which relies on this? If not then I think we could, if we're not sure I would lean to keeping it to avoid breaking anyone?

DomAyre: We can probably remove it eventually either way

beejones: no this was just for setting the first key in akv. If we delete it we need to change docs to point to the new scripts.

DomAyre: Happy to remove then if nothing uses it
https://github.com/microsoft/azure-privacy-sandbox-kms/pull/240/commits/eb4a4a5a6ada3b7b3525293e0f01ffab4c075798

beejones: please update docs to points to the new script

DomAyre: I did its in the commit mentioned above

beejones: docs/Create_new_akv_member_key.md still references create_member_cert_akv

DomAyre: That file is deleted in the change I mentioned above (specifically I renamed it to `Adding_members.md` since it applies to both AKV and non AKV cases, and git sees it as deleting the old file and replacing with the new one)

scripts/akv/key_policy.json
```diff
@@ -0,0 +1,30 @@
+{
+    "issuerParameters": {
```
takuro-sato, line 2: Just a question, but who defines the json schema?

scripts/ccf/member/add.sh
```diff
@@ -0,0 +1,87 @@
...
+            --vault-name $AKV_VAULT_NAME \
+            --name $MEMBER_NAME \
+            --file $WORKSPACE/${MEMBER_NAME}_cert.pem
+    fi
+
+    # Propose adding the member to the network
+    MEMBER_CERT=$(jq -Rs . < $WORKSPACE/${MEMBER_NAME}_cert.pem) \
+        envsubst < $REPO_ROOT/governance/proposals/set_member.json | jq \
+            > $WORKSPACE/proposals/set_member_${MEMBER_NAME}.json
+    ccf-propose $WORKSPACE/proposals/set_member_${MEMBER_NAME}.json
```
takuro-sato, line 47: Who provides the command ccf-propose?

scripts/akv/key_policy.json
takuro-sato: Found that the cleanroom tool consumes it

scripts/ccf/member/add.sh
```diff
@@ -0,0 +1,87 @@
...
+            "Proposal submitted, await sufficient votes for members to be accepted\n" \
+            "Check member status with:\n" \
+            "$REPO_ROOT/scripts/ccf/member/info.sh"
+            return 0
+        else
+            return 1
+        fi
+    fi
+
+    (
```
takuro-sato, line 62: For what reason do we want subshell here?

scripts/ccf/member/info.sh
```diff
@@ -0,0 +1,62 @@
...
+    export MEMBER_ID=`ccf-member-id $MEMBER_NAME`
+    if_active_member=`curl -s $KMS_URL/gov/service/members?api-version=2024-07-01 \
+        --cacert $KMS_SERVICE_CERT_PATH \
+        | jq -e ".value[] | select(.memberId==\"$MEMBER_ID\")"`
+
+    if [ -n "$if_active_member" ]; then
+        echo $if_active_member
+        return
+    fi
+    export MEMBER_CERT=$(cat $WORKSPACE/${MEMBER_NAME}_cert.pem)
```
takuro-sato, line 28: Is it possible to do `export` later to make the interface clearer?
I'm thinking about having lines at the bottom of the function like:
```
export MEMBER_NAME
export MEMBER_CERT
...
```

This line can be just `MEMBER_CERT=$(cat $WORKSPACE/${MEMBER_NAME}_cert.pem)` without export.

scripts/ccf/member/add.sh
takuro-sato: Found the answer. It's sourced from `source $REPO_ROOT/scripts/ccf/propose.sh`.

takuro-sato (commented):
Looks good. Just left minor comments

scripts/akv/key_policy.json
DomAyre: I don't think cleanroom consumes it unless I'm forgetting something?

I presume the policy is defined by AKV or some standard, it describes the member credentials

scripts/ccf/member/info.sh
DomAyre: I think the tricky thing with that scheme is there are multiple exit points so it wouldn't allow us to move the exporting much lower down, not sure the benefits are worth it?

scripts/ccf/member/add.sh
DomAyre: I had to remind myself so definitely worth a comment.

Added in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/240/commits/637f2e21c90d6f8da5a34184dd142afb1ac97ce8

scripts/ccf/member/info.sh
takuro-sato: Ah, that's true. I agree with you.

takuro-sato: Maybe it might be worth thinking about moving these script to something like python which can handle states with class easily (i.e. just by following conventional ways) , but I guess not now

takuro-sato (approved):

PR #242, opened 2024-12-06
Make setting constitution more flexible

DomAyre:
### Motivation
This is a prerequisite to properly testing #240. As we would like to test adding members both when they are immediately accepted and when voting is required

### Changes
- [x] Structure the `governance/constitution` directory to have separate scripts for the bits of the CCF constitution, with options where appropriate (e.g. type of resolve function)
- [x] Re-write `constitution_set.sh` to allow the user to specify which parts of the constitution to use
- [x] Update usage of this script to maintain existing behaviour (i.e. auto accept proposals and use kms specific actions)
- [x] Rename endpoint tests to just system test (singular) because we now test more than just endpoints
- [x] Add a `test_constitution.py` test which tests the different resolve options we have


.github/workflows/system-tests.yml
```diff
@@ -24,14 +24,14 @@ jobs:
           JSON_ARRAY=$(printf '%s\n' "${TEST_FILES[@]}" | jq -R . | jq -s .)
           echo "tests=$JSON_ARRAY" | sed ':a;N;$!ba;s/\n//g' >> $GITHUB_OUTPUT
 
-  endpoint:
-    name: endpoint
+  test:
+    name: ${{ '' }}
```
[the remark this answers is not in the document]
DomAyre: It is intended

governance/constitution/actions/default.js
```diff
@@ -0,0 +1,1481 @@
...
+        info.state = "Dropped";
+        proposalsMap.set(k, ccf.jsonCompatibleToBuf(info));
+      }
+    }
+  });
+}
+
+function setServiceCertificateValidityPeriod(validFrom, validityPeriodDays) {
+  const rawConfig =
+    ccf.kv["public:ccf.gov.service.config"].get(getSingletonKvKey());
```
beejones, line 170: is the key going to be all zero?

governance/constitution/actions/default.js
```diff
@@ -0,0 +1,1481 @@
...
+  if (rawConfig === undefined) {
+    throw new Error("Service configuration could not be found");
+  }
+  const serviceConfig = ccf.bufToJsonCompatible(rawConfig);
+
+  const default_validity_period_days = 365;
```
beejones, line 176: how are these certs rolled?

governance/constitution/actions/default.js
```diff
@@ -0,0 +1,1481 @@
...
+
+  const serviceInfoTable = "public:ccf.gov.service.info";
+  const rawServiceInfo = ccf.kv[serviceInfoTable].get(getSingletonKvKey());
+  if (rawServiceInfo === undefined) {
+    throw new Error("Service info could not be found");
+  }
+  const serviceInfo = ccf.bufToJsonCompatible(rawServiceInfo);
+
+  serviceInfo.cert = renewed_service_certificate;
+  ccf.kv[serviceInfoTable].set(
```
beejones, line 203: you could have use the existing settings map

governance/constitution/actions/default.js
```diff
@@ -0,0 +1,1481 @@
...
+  nodeInfo,
+  validFrom,
+  validityPeriodDays,
+) {
+  if (nodeInfo.certificate_signing_request === undefined) {
+    throw new Error(`Node ${nodeId} has no certificate signing request`);
+  }
+
+  const rawConfig =
+    ccf.kv["public:ccf.gov.service.config"].get(getSingletonKvKey());
```
beejones, line 220: same here. We could save all settings in the same map.

governance/constitution/actions/default.js
```diff
@@ -0,0 +1,1481 @@
...
+  if (rawConfig === undefined) {
+    throw new Error("Service configuration could not be found");
+  }
+  const serviceConfig = ccf.bufToJsonCompatible(rawConfig);
+
+  const default_validity_period_days = 365;
```
beejones, line 226: this value is saved in your serviceInfoTable. Should we use the same value?

governance/constitution/actions/default.js
```diff
@@ -0,0 +1,1481 @@
+class Action {
```
beejones, line 1: is this the default action as it is defined in CCF?
If so,  why do we redefine it. Our current set_constitution just specified the delta we define in KMS.

test/system-test/test_constitution.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+import tempfile
+
+from utils import apply_kms_constitution, propose, get_test_action, get_test_proposal
+
+def test_constitution_auto_accept(setup_kms):
+    with get_test_action() as action_file:
+        apply_kms_constitution(actions=action_file.name)
```
beejones, line 8: do you have multiple nodes voting in this test?

governance/constitution/actions/default.js
DomAyre: Because baking the assumption into the script that we will add stuff on top of the sandbox default constitution makes changing things difficult. (What if we want a different resolve function?)

test/system-test/test_constitution.py
DomAyre: We don't have any voting in this test, correct me if I'm wrong but there's no standalone scripts for voting in the KMS, it's all baked into scripts doing other things mainly proposing.

I add scripts in the next PR since we will want tests for it, but in this PR the change is just about setting constitutions and showing that a proposal is open is enough to demonstrate whether the resolve function auto accepts or not

governance/constitution/actions/default.js
```diff
@@ -0,0 +1,1481 @@
...
+}
+
+function checkJwks(value, field) {
+  checkType(value, "object", field);
+  checkType(value.keys, "array", `${field}.keys`);
+  for (const [i, jwk] of value.keys.entries()) {
+    checkType(jwk.kid, "string", `${field}.keys[${i}].kid`);
+    checkType(jwk.kty, "string", `${field}.keys[${i}].kty`);
+    checkType(jwk.x5c, "array", `${field}.keys[${i}].x5c`);
+    checkLength(jwk.x5c, 1, null, `${field}.keys[${i}].x5c`);
```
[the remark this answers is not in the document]
DomAyre: See https://github.com/microsoft/azure-privacy-sandbox-kms/pull/242#issuecomment-2532277071

governance/constitution/actions/default.js
DomAyre: See https://github.com/microsoft/azure-privacy-sandbox-kms/pull/242#issuecomment-2532277071

governance/constitution/actions/default.js
DomAyre: See https://github.com/microsoft/azure-privacy-sandbox-kms/pull/242#issuecomment-2532277071

governance/constitution/actions/default.js
DomAyre: See https://github.com/microsoft/azure-privacy-sandbox-kms/pull/242#issuecomment-2532277071

beejones (approved):

governance/constitution/actions/default.js
beejones: you can add just your delta's to the default constitution. That is what I did with kms_sctions.js.
Now you will have to maintain the constitution, changed in CCF have to be copied.

beejones:
I suggest to avoid copying the default constitution files and  just have the delta;s to avoid having to maintain these files.

DomAyre:
> https://github.com/microsoft/azure-privacy-sandbox-kms/pull/242#issuecomment-2536153691
  I suggest to avoid copying the default constitution files and just have the delta;s to avoid having to maintain these files.
  
Can you explain how to specify a delta of a resolve function which isn't just a patch against the default one? A patch has the same maintenance overhead as the actual code but with the code it refers to being invisible.

governance/constitution/actions/default.js
DomAyre: The existing approach is fine if you're just appending to the existing constitution, but if you want to change something in it (you suggested many possible changes, I've implemented different resolve functions as a concrete example) then your approach still requires committed code, likely a patch? 

If that code is indeed a patch against assumed defaults, it will need maintenance if the default constitution changes, in fact your suggested approach would break whereas the one in this PR can be updated if we choose but will continue to work if we don't

PR #259, opened 2024-12-10
Add support and testing for scaling nodes

DomAyre:
### Motivation
Part of proving out `az-cleanroom`, but also useful for showing the KMS in more production like environments

### Changes
- [x] Add `scale-nodes.sh` scripts to both test environments which scale the number of nodes in the ccf network to a specified number
- [x] Added a new test file which tests scaling nodes up and down
- [x] Prevent local sandbox nodes from stepping on each others workspaces


scripts/ccf/az-cleanroom-aci/up.sh
```diff
@@ -43,6 +43,8 @@ az-cleanroom-aci-up() {
     sudo cp $KMS_SERVICE_CERT_PATH /usr/local/share/ca-certificates/kms_ca.crt
```
beejones, line 43: cool. So we no longer have to pass service_cert in curl

scripts/ccf/sandbox_local/down.sh
```diff
@@ -7,7 +7,7 @@ ccf-sandbox-local-down() {
     set -e
 
     REPO_ROOT="$(realpath "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../../..")"
-    docker compose -f $REPO_ROOT/services/docker-compose.yml down ccf-sandbox --remove-orphans
+    docker compose -f $REPO_ROOT/services/docker-compose.yml down --remove-orphans
 
     unset KMS_URL
```
beejones, line 12: why do we unset env vars when stopping containers? They will be set in the up commands.

scripts/ccf/sandbox_local/join_config.json
```diff
@@ -0,0 +1,70 @@
...
+      "service_certificate_file": "/workspace/sandbox_common/service_cert.pem",
+      "join": {
+        "retry_timeout": "1s",
+        "target_rpc_address": "127.0.0.1:8000",
+        "follow_redirect": true
+      }
+    },
+    "output_files": {
+      "node_certificate_file": "${NODE_WORKSPACE}/0.pem",
+      "pid_file": "${NODE_WORKSPACE}/0.pid",
```
beejones, line 66: don't we need to set the CN= for the certs?

scripts/ccf/sandbox_local/scale-nodes.sh
```diff
@@ -0,0 +1,140 @@
...
+            proposal_file="$WORKSPACE/proposals/remove_${node_id}.json"
+            NODE_ID=$node_id \
+                envsubst < "$REPO_ROOT/governance/proposals/remove_node.json" \
+                    | jq > "$proposal_file"
+            ccf-propose "$proposal_file"
+        done
+    fi
+
+    # Wait until the node count matches the desired value
+    retries=10
```
beejones, line 119: will 10 secs be deterministic? I assume it will depend on the number of nodes.

scripts/ccf/sandbox_local/join_config.json
```diff
@@ -0,0 +1,70 @@
...
+      "type": "Join",
+      "service_certificate_file": "/workspace/sandbox_common/service_cert.pem",
+      "join": {
+        "retry_timeout": "1s",
+        "target_rpc_address": "127.0.0.1:8000",
+        "follow_redirect": true
+      }
+    },
+    "output_files": {
+      "node_certificate_file": "${NODE_WORKSPACE}/0.pem",
```
beejones, line 65: are we using the same file structure as sandbox?

scripts/ccf/sandbox_local/scale-nodes.sh
```diff
@@ -0,0 +1,140 @@
...
+        --cacert $KMS_SERVICE_CERT_PATH | jq '.nodes'
+}
+
+ccf-sandbox-local-node() {
+    curl -s $KMS_URL/node/network/nodes/$1 \
+        --cacert $KMS_SERVICE_CERT_PATH | jq
+}
+
+status-is() {
+    jq --arg status "$1" '[.[] | select(.status == $status)]'
```
beejones, line 17: what is this argument? Maybe good to add a comment

scripts/kms/constitution_set.sh
```diff
@@ -10,7 +10,8 @@ constitution-set() {
   CONSTITUTION_PATH=$1
```
beejones, line 10: can we test if $1 was passed? Will also be clearer to understand the scripts

scripts/ccf/az-cleanroom-aci/up.sh
DomAyre: You raise a good point, I had a think about it and made a separate PR since it's unrelated to this change.
https://github.com/microsoft/azure-privacy-sandbox-kms/pull/264

scripts/ccf/sandbox_local/down.sh
DomAyre: The environment variables that are set when we bring up a service describe that instance of that service, when we bring that service down those variables are no longer applicable because the instance of that service no longer exists, so in order to avoid mistakes we should unset them.

scripts/ccf/sandbox_local/join_config.json
DomAyre: I don't know, do we? The sandbox.sh script appears to set arbitrary values for these so I didn't think it mattered?

scripts/ccf/sandbox_local/scale-nodes.sh
DomAyre: It's a good question, and thinking about it I realised for non auto-accept resolve function constitutions, this will always fail since the trust proposal won't be accepted until it's voted on.

So in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/259/commits/38892485ef3cb488badfd8635bbe0723c72cb488 I replace this with a loop which will print outstanding proposals so the users can then easily vote on them

scripts/ccf/sandbox_local/join_config.json
DomAyre: Yes where it matters
```
workspace/
├── join_config.json
├── node-b621a4a8-8fa7-40dc-8451-83ee87644753
│   ├── 0.node_address
│   ├── 0.pem
│   ├── 0.pid
│   ├── 0.rpc_addresses
│   ├── err
│   └── out
├── proposals
│   └── trust_0b374cca774ad0322ef0141fdbf07cdd66b59614b6bb1281da2651ee98a43cce.json
├── sandbox_0
│   ├── 0.config.json -> /workspace/sandbox_common/0.config.json
│   ├── 0.ledger
│   │   ├── ledger_1-10.committed
│   │   └── ledger_11
│   ├── 0.node_address
│   ├── 0.pem
│   ├── 0.rpc_addresses
│   ├── 0.snapshots
│   │   └── snapshot_10_11.committed
│   ├── actions.js
│   ├── apply.js
│   ├── cchost -> /opt/ccf_virtual/bin/cchost
│   ├── err
│   ├── libjs_generic.virtual.so -> /opt/ccf_virtual/lib/libjs_generic.virtual.so
│   ├── member0_cert.pem
│   ├── member0_enc_pubk.pem
│   ├── node.pid
│   ├── out
│   ├── resolve.js
│   ├── service_cert.pem
│   └── validate.js
└── sandbox_common
    ├── 0.config.json
    ├── 0.node_address
    ├── 0.pem
    ├── 0.rpc_addresses
    ├── actions.js
    ├── apply.js
    ├── keygenerator.sh -> /opt/ccf_virtual/bin/keygenerator.sh
    ├── member0_cert.pem
    ├── member0_enc_privk.pem
    ├── member0_enc_pubk.pem
    ├── member0_privk.pem
    ├── resolve.js
    ├── service_cert.pem
    ├── set_js_app_proposal.json
    ├── set_js_app_vote_for.json
    ├── user0_cert.pem
    ├── user0_privk.pem
    └── validate.js
```

scripts/ccf/sandbox_local/scale-nodes.sh
DomAyre: Sorry what do you mean? You're wondering what the positional argument to `status-is` is? If so, isn't it obvious since the function is called `status-is` and it's immediately compared to the `.status` field to select elements?

scripts/kms/constitution_set.sh
DomAyre: This file is completely changed in #242 so I don't think it's worth changing here especially since it's unrelated to the change in this PR

DomAyre:
CI currently fails because in order to support submitting proposals to the non-primary (which can happen now we have multiple nodes) I removed http2 interfaces which currently breaks e2e testing follow #263 for resolution

scripts/ccf/sandbox_local/join_config.json
beejones: I believe it matters for using curl. @takure-sato request mCCF to add the CN=.. property to the x.509 cert for TLS termination. We need to check with Takuro.

DomAyre: This isn't anything to do with mCCF though? This is a network running purely with local containers running the CCF sandbox?

DomAyre:
To remove the dependency on removing HTTP2, we can just manually find the primary and send proposals directly to it

beejones (approved):

PR #260, opened 2024-12-11
Fix and test adding new issuers 

beejones:
Fixed the bug that a new issuer was not found.
Added auth tests
Test adding a new issuer
Test calling with a wrong issuer
Changed our test issuer to access properties to be put into the token


scripts/ccf/az-cleanroom-aci/up.sh
```diff
@@ -39,6 +39,10 @@ az-cleanroom-aci-up() {
     export KMS_SERVICE_CERT_PATH="$WORKSPACE/service_cert.pem"
     export KMS_MEMBER_CERT_PATH="$WORKSPACE/ccf-operator_cert.pem"
     export KMS_MEMBER_PRIVK_PATH="$WORKSPACE/ccf-operator_privk.pem"
+    export KMS_USER_CERT_PATH="$WORKSPACE/sandbox_common/user0_cert.pem"
+    export KMS_USER_PRIVK_PATH="$WORKSPACE/sandbox_common/user0_privk.pem"
+    export JWT_TOKEN_ISSUER_URL="http://localhost:3000/token"
```
DomAyre, line 44: JWT Issuer details shouldn't be set here this is just about bringing up CCF. We already set them when bringing up the test issuer

scripts/jwt_issuer/up.sh
```diff
@@ -23,5 +24,5 @@ jwt-issuer-up
 
 jq -n '{
     JWT_ISSUER_WORKSPACE: env.JWT_ISSUER_WORKSPACE,
-    JWT_ISSUER: env.JWT_ISSUER,
```
DomAyre, line 26: Why have we removed JWT_ISSUER? The idea of this part is anything we've exported and is therefore important to future execution is printed here in case the environment can't be persisted easily, namely in the system tests

scripts/jwt_issuer/up.sh
```diff
@@ -14,7 +14,8 @@ jwt-issuer-up() {
 
     sudo chown $USER:$USER -R $JWT_ISSUER_WORKSPACE
 
-    export JWT_ISSUER="http://localhost:3000/token"
+    export JWT_TOKEN_ISSUER_URL="http://localhost:3000/token"
```
DomAyre, line 17: Can you explain the difference between these two things? They're both URLs and I presume they both issue tokens so maybe we can think of a better naming scheme?

test/system-test/test_auth.py
```diff
@@ -0,0 +1,47 @@
+from endpoints import auth
+from utils import apply_kms_constitution, trust_jwt_issuer
+
+
+def test_auth_member_cert(setup_kms):
+    status_code, auth_json = auth(auth="member_cert")
+    print(auth_json)
```
DomAyre, line 7: I like this change to print the output of the endpoint call for tests it makes the output much clearer what happened, maybe we can put this print inside the call_endpoint wrapper so all tests benefit?

scripts/kms/jwt_issuer_trust.sh
```diff
@@ -8,6 +8,24 @@ jwt-issuer-trust() {
 
   REPO_ROOT="$(realpath "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../..")"
 
+  # Check for new issuer
```
DomAyre, line 11: For my own understanding, the scheme is:

- If you call this without --iss it will use JWT_TOKEN_ISSUER_URL which will have been set if you bring up a local test JWT issuer
- If you have some external JWT issuer you can just specify it with this --iss flag?

If my understanding is correct I really like this scheme!

scripts/jwt_issuer/up.sh
DomAyre: Also not super keen on JWT_TOKEN because the T in JWT stands for token.

So we have JSON Web Token Token Issuer

test/system-test/test_auth.py
beejones: This is already outputted in endpoints

scripts/jwt_issuer/up.sh
beejones: JWT_ISSUER is the name used in CCF for registering the JWT issuer.
JWT_TOKEN_ISSUER_URL is the actual url of the endpoint.

beejones: it is very common to talk about JWT tokens. Just do a search for "JWT token".

scripts/kms/jwt_issuer_trust.sh
beejones: For an external one we need to set JWT_TOKEN_ISSUER_URL  to the token endpoint.
--iss will the name for the iss property in the token. CCF uses this to register the issuer.

scripts/jwt_issuer/up.sh
DomAyre: > JWT_ISSUER is the name used in CCF for registering the JWT issuer. JWT_TOKEN_ISSUER_URL is the actual url of the endpoint.

Ah thanks for the explanation, that makes sense

> it is very common to talk about JWT tokens. Just do a search for "JWT token".

Being common doesn't make it correct or less redundant, and I wouldn't claim it's more common that just JWT, [jwt.io](https://jwt.io) always just says JWT.

How about we use 
- `JWT_ISSUER_URL` which is the URL which issues JWTs
- `JWT_ISSUER_NAME` which is the name of the thing which issues JWTs (for CCF)
  - We could make it clearly a name by making it not look like a URL?

test/system-test/test_auth.py
DomAyre: Ah awesome, so why are we printing it a second time just after it's been printed?

scripts/kms/jwt_issuer_trust.sh
DomAyre: So is it equivalent to setting `JWT_ISSUER`?  In which case why are we mixing methods? I think we either allow fully specifying which JWT issuer we're trusting via command line params or we just require the env variables to be changed?

scripts/ccf/sandbox_local/up.sh
```diff
@@ -18,6 +18,10 @@ ccf-sandbox-local-up() {
     export KMS_SERVICE_CERT_PATH="$WORKSPACE/sandbox_common/service_cert.pem"
     export KMS_MEMBER_CERT_PATH="$WORKSPACE/sandbox_common/member0_cert.pem"
     export KMS_MEMBER_PRIVK_PATH="$WORKSPACE/sandbox_common/member0_privk.pem"
+    export KMS_USER_CERT_PATH="$WORKSPACE/sandbox_common/user0_cert.pem"
+    export KMS_USER_PRIVK_PATH="$WORKSPACE/sandbox_common/user0_privk.pem"
+    export JWT_TOKEN_ISSUER_URL="http://localhost:3000/token"
```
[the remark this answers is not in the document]
DomAyre: We shouldn't set user certs here, it's an implementation detail of sandbox.sh that a user is created and added to the network, not all ccf networks will have a pre-created user. We should have a script to add new users like we do for members then it's agnostic to how the network was deployed

scripts/ccf/az-cleanroom-aci/up.sh
DomAyre: The comment https://github.com/microsoft/azure-privacy-sandbox-kms/pull/260#discussion_r1882471375 also applies here

scripts/ccf/sandbox_local/up.sh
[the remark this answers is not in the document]
beejones: I will remove user certs for now

scripts/jwt_issuer/up.sh
beejones: added it back

scripts/ccf/az-cleanroom-aci/up.sh
beejones: removed

scripts/jwt_issuer/up.sh
beejones: I believe CCF requires this to be a url. Need to check this.

scripts/kms/jwt_issuer_trust.sh
beejones: I added a test adding a new issuer implemented by our test token issuer.
For this case I needed to specify the issuer JWT_ISSUER and keep the url to the test token url.

beejones:
tests:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/12313566584

scripts/kms/jwt_issuer_trust.sh
DomAyre: I'm not sure that addresses my point about mixed methods, for the test you could set both env variables for the test, or you could provider command line params for both and specify them for the test?

scripts/jwt_issuer/up.sh
DomAyre: If it needs to be a URL, could it just be the actual URL so we only need one variable instead of two?

scripts/ccf/az-cleanroom-aci/up.sh
DomAyre: Great, my original comment about JWT issuer information still applies

scripts/ccf/sandbox_local/up.sh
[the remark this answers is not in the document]
DomAyre: Thanks, my original comment about JWT issuer info still applies

scripts/jwt_issuer/up.sh
```diff
@@ -14,8 +14,6 @@ jwt-issuer-up() {
 
     sudo chown $USER:$USER -R $JWT_ISSUER_WORKSPACE
 
-    export JWT_TOKEN_ISSUER_URL="http://localhost:3000/token"
```
DomAyre: Why have you removed this? We want to export the details of the JWT issuer when we bring it up? I think you've mixed it up with my comment that JWT issuer info shouldn't be exported when we bring up CCF

scripts/jwt_issuer/up.sh
```diff
@@ -24,5 +22,4 @@ jwt-issuer-up
 
 jq -n '{
     JWT_ISSUER_WORKSPACE: env.JWT_ISSUER_WORKSPACE,
-    JWT_TOKEN_ISSUER_URL: env.JWT_TOKEN_ISSUER_URL,
```
DomAyre: Why have you removed this?

scripts/jwt_issuer/up.sh
DomAyre: I still don't see it?

scripts/jwt_issuer/up.sh
beejones: indeed.
{
  bool validate_issuer(
    const std::string& iss,
    const std::optional<std::string>& tid,
    std::string constraint)
  {
    LOG_DEBUG_FMT(
      "Verify token.iss {} and token.tid {} against published key issuer {}",
      iss,
      tid,
      constraint);

    const auto issuer_url = ::http::parse_url_full(constraint);

beejones: > If it needs to be a URL, could it just be the actual URL so we only need one variable instead of two?

See https://github.com/microsoft/azure-privacy-sandbox-kms/pull/260#discussion_r1883672015

DomAyre:
> tests: https://github.com/microsoft/privacy-sandbox-dev/actions/runs/12313566584

@beejones I think the privacy-sandbox-dev CI is currently broken, I'm looking into it now and I'll update when it's fixed

scripts/jwt_issuer/up.sh
beejones: fixed

scripts/jwt_issuer/up.sh
beejones: fixed

DomAyre:
Now privacy-sandbox-dev CI is fixed, here is a run for this PR
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/12369513461

scripts/ccf/az-cleanroom-aci/up.sh
DomAyre: This isn't resolved

scripts/ccf/sandbox_local/up.sh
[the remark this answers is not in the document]
DomAyre: This isn't resolved

DomAyre:
Merged the fixes in main to this PR, re-running privacy-sandbox-dev CI
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/12370752558

test/utils/jwt/README.md
```diff
@@ -0,0 +1,9 @@
+# Token issuer
+
+## Request
+```
+curl -X POST ${AadEndpoint:-http://localhost:3000/token} \
```
DomAyre, line 5: This needs updating since #266

scripts/jwt_issuer/up.sh
```diff
@@ -9,11 +9,13 @@ jwt-issuer-up() {
     REPO_ROOT="$(realpath "$(dirname "$(realpath "${BASH_SOURCE[0]}")")/../..")"
     export JWT_ISSUER_WORKSPACE=${JWT_ISSUER_WORKSPACE:-$REPO_ROOT/jwt_issuers_workspace/${UNIQUE_ID:-default}/}
     mkdir -p $JWT_ISSUER_WORKSPACE
+    export JWT_TOKEN_ISSUER_URL="http://localhost:3000/token"
+    export JWT_ISSUER="http://Demo-jwt-issuer"
```
beejones, line 13: only use JWT_ISSUER

scripts/jwt_issuer/down.sh
```diff
@@ -9,3 +9,4 @@ docker compose -p ${UNIQUE_ID:-default} -f services/docker-compose.yml down jwt-
 
 unset JWT_ISSUER_WORKSPACE
 unset JWT_ISSUER
+unset JWT_TOKEN_ISSUER_URL
```
DomAyre, line 12: We will update this too when we combine JWT_ISSUER into one value

scripts/kms/endpoints/auth.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+set -ex
+
```
beejones, line 7: remove +x

scripts/kms/js_app_set.sh
```diff
@@ -22,6 +22,10 @@ js-app-set() {
       echo "$line" >> "$temp_file"
     fi
   done < $REPO_ROOT/governance/proposals/set_js_app.json
+
+  # Ensure the target directory exists
+  mkdir -p $WORKSPACE/proposals
```
DomAyre, line 27: We can revisit this when all other changes are addressed, this shouldn't be necessary and it might hide other issues

scripts/kms/jwt_issuer_trust.sh
DomAyre: Now we've decided to unify JWT_ISSUER this goes away

test/system-test/conftest.py
```diff
@@ -55,7 +55,7 @@ def setup_jwt_issuer():
             ["./scripts/jwt_issuer/up.sh", "--build"],
             env={
                 **os.environ,
-                "JWT_ISSUER_WORKSPACE": f"{REPO_ROOT}/jwt_issuer_workspace",
+                "JWT_ISSUER_WORKSPACE": f"{REPO_ROOT}/jwt_issuers_workspace",
```
DomAyre, line 58: We can match the default in up.sh (/defualt/)

PR #263, opened 2024-12-12
Remove usage of HTTP2

DomAyre:
### Motivation
I noticed that if the primary node of the CCF network changes, or if there is a load balancer which might not point at the primary, then proposals will fail because HTTP2 nodes cannot forward proposals to the primary.

```
{
  "error": {
    "code": "NotImplemented",
    "message": "Request cannot be forwarded to primary on HTTP/2 interface."
  }
}
```

I don't believe there is any issue with dropping HTTP2, but I presume it was added for a reason so I'm raising a standalone PR in order to discuss.

### Update

Looks like the requirement on HTTP2 comes from the e2e tests, I don't think our testing framework should force us to use an older HTTP version so I suggest updating the E2E tests, or if the coverage is similar to System Tests maybe we can remove them?


beejones:
> ### Motivation
> I noticed that if the primary node of the CCF network changes, or if there is a load balancer which might not point at the primary, then proposals will fail because HTTP2 nodes cannot forward proposals to the primary.
> 
> ```
> {
>   "error": {
>     "code": "NotImplemented",
>     "message": "Request cannot be forwarded to primary on HTTP/2 interface."
>   }
> }
> ```
> 
> I don't believe there is any issue with dropping HTTP2, but I presume it was added for a reason so I'm raising a standalone PR in order to discuss.
> 
> ### Update
> Looks like the requirement on HTTP2 comes from the e2e tests, I don't think our testing framework should force us to use an older HTTP version so I suggest updating the E2E tests, or if the coverage is similar to System Tests maybe we can remove them?

I remember that @akuro-sato added HTTP2 for compatibility with B&A services. We need to check the compatibility with this change.

DomAyre:
Tested privacy-sandbox-dev with this branch
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/12299432676

PR #266, opened 2024-12-17
Allow running concurrent JWT issuers

DomAyre:
### Motivation

If we want to run multiple concurrent JWT issuers for testing in #260 we need to be able to have separate instances of the test JWT issuer run at once

### Changes
- [x] Define a `UNIQUE_ID` environment variable to delimit iss
- [x] Scope docker compose invocations for JWT issuer to a project name based on `UNIQUE_ID`
- [x] Update test JWT issuer implementation to bind to a random available port and emit the address in a file

### DIFF
--- a/services/docker-compose.yml
+++ b/services/docker-compose.yml
@@ -24,4 +24,4 @@ services:
       interval: 1s
       retries: 120
     volumes:
-      - ${JWT_ISSUER_WORKSPACE:-../jwt_issuer_workspace}:/workspace
+      - ${JWT_ISSUER_WORKSPACE:-../jwt_issuers_workspace/default}:/workspace
--- a/services/docker-compose.yml
+++ b/services/docker-compose.yml
@@ -20,17 +20,8 @@ services:
     environment:
       - KMS_WORKSPACE=/workspace
     healthcheck:
-      test:
-        [
-          "CMD",
-          "curl",
-          "-k",
-          "--fail",
-          "-X",
-          "POST",
-          "http://localhost:3000/token",
-        ]
+      test: "curl -k --fail -X POST $(cat /workspace/jwt_issuer_address)/token"
       interval: 1s
       retries: 120
     volumes:
-      - ${JWT_ISSUER_WORKSPACE:-./}:/workspace
+      - ${JWT_ISSUER_WORKSPACE:-../jwt_issuer_workspace}:/workspace
--- a/test/utils/jwt/src/index.ts
+++ b/test/utils/jwt/src/index.ts
@@ -12,8 +12,7 @@ const proposalsPath = `${process.env.KMS_WORKSPACE}/proposals`;
 const privateKeyPath = `${process.env.KMS_WORKSPACE}/private.pem`;
 const certificatePath = `${process.env.KMS_WORKSPACE}/cert.pem`;
 const kid = "Demo IDP kid";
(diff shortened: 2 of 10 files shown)


test/utils/jwt/src/index.ts
```diff
@@ -12,8 +12,7 @@ const proposalsPath = `${process.env.KMS_WORKSPACE}/proposals`;
 const privateKeyPath = `${process.env.KMS_WORKSPACE}/private.pem`;
 const certificatePath = `${process.env.KMS_WORKSPACE}/cert.pem`;
 const kid = "Demo IDP kid";
-const hostPort = 3000;
-const host = `http://localhost:${hostPort}`;
+const host = `http://localhost:0`;
```
beejones, line 15: with a fixed host, how will CCF distinguish between different issuers:
await Proposals.issuerProposal(host, proposalsPath);

We will need to add the the unique id because otherwise different proposals will just update the existing proposal, or is this intentional?

services/docker-compose.yml
```diff
@@ -20,17 +20,8 @@ services:
     environment:
       - KMS_WORKSPACE=/workspace
```
beejones, line 21: shouldn't we expose the random port used by the jwt issuer?

test/utils/jwt/src/index.ts
DomAyre: I don't fully understand your question, we've moved **from** a fixed host **to** a dynamic one in this change. Regarding CCF, how we register the issuer is a topic of discussion in your other PR, my preference as discussed [here](https://github.com/microsoft/azure-privacy-sandbox-kms/pull/260#discussion_r1886607870) is that we just pass the address of the issuer which will as of this change be unique per issuer (at least locally)

services/docker-compose.yml
DomAyre: We should and we do, otherwise it wouldn't be possible to get tokens, if you look two lines down you'll see it's emitted into a file in the workspace which we can read

test/utils/jwt/src/index.ts
beejones: Let me clarify. When you register two issuers as localhost:0, for CCF this is the same issuer. Your two issuers have different keys, so validation of token from the first issuer will fail because CCF will use the key of the second to validate

DomAyre: We won't register it as localhost:0, we'll use the actual address including the resolved port

beejones: in index,ts localhost:0 is registered as issuer for all issuers
// generate issuer proposal
    await Proposals.issuerProposal(host, proposalsPath);

beejones: > in index,ts localhost:0 is registered as issuer for all issuers // generate issuer proposal await Proposals.issuerProposal(host, proposalsPath);

Dom and I agreed this line needs to be removed because it is no longer needed. The registration of issuers is done with the correct port in scripts. Dom will do a seperate PR to remove the line

beejones (approved):

test/utils/jwt/src/index.ts
DomAyre: Here's the separate PR to clean up old code:

https://github.com/microsoft/azure-privacy-sandbox-kms/pull/283

PR #274, opened 2025-01-06
Roll back to ubuntu 22.04

DomAyre:
Looks like the CI fails consistently since using ubuntu 24.04, so rolling back to 22.04 to see if it improves things

### DIFF
--- a/.github/workflows/system-test.yml
+++ b/.github/workflows/system-test.yml
@@ -18,7 +18,7 @@ jobs:
 
   test:
     name: ${{ inputs.test }} (${{ matrix.config.env }}) (${{ matrix.config.use_akv && 'akv' || 'local' }}_keys)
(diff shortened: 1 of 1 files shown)


DomAyre:
I think this should still be merged despite the e2e test failure, as it fixes the catastrophic akv_key and az-cleanroom consistent failures and we saw the e2e failure intermittently before the main issue

DomAyre:
E2E test flakiness addressed in #276

KenGordon (approved):
As discussed, approved pending an issue to be raised describing our guesses about why using ubuntu 24.04 rather than 22.04 may lead to delays and so the token expiring. In particular that with 22.04 the "install dependencies" step takes 42 seconds whereas with 24.04 it takes minutes. The token only lasts for 5 minutes.

PR #276, opened 2025-01-06
Use ts-node to run e2e tests

DomAyre:
### Motivation

The errors we see intermittently in the e2e tests have this message
```
Error: or [Error]: Unknown file extension ".ts" for /workspaces/azure-privacy-sandbox-kms/test/e2e-test/src/index.ts
```

It appears that the current method of running typescript files directly with node is very error prone, so using ts-node instead

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
   workflow_dispatch:
   schedule:
     - cron: "0 1 * * *"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -8,7 +8,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
```
beejones, line 11: Not important. Just in case you are not aware. This change was done is https://github.com/microsoft/azure-privacy-sandbox-kms/pull/275/files also

beejones (approved):

.github/workflows/ci.yml
DomAyre: Don't worry I'm aware, I've done it in any PR that is based on #274 so that getting a CI run isn't blocked on that merging

PR #277, opened 2025-01-06
Add a retry mechanism to scaling nodes in ACI

DomAyre:
### Motivation
A source of flakiness I have identified is when we attempt to scale the number of nodes in ACI. To alleviate this, I've added a retry mechanism

### DIFF
--- a/.github/workflows/system-test.yml
+++ b/.github/workflows/system-test.yml
@@ -18,7 +18,7 @@ jobs:
 
   test:
     name: ${{ inputs.test }} (${{ matrix.config.env }}) (${{ matrix.config.use_akv && 'akv' || 'local' }}_keys)
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/system-test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
 
   test:
     name: ${{ inputs.test }} (${{ matrix.config.env }}) (${{ matrix.config.use_akv && 'akv' || 'local' }}_keys)
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
beejones, line 21: just curious.
Was the latest version a clear source of failing or are you just prudent?

beejones (approved):

.github/workflows/system-test.yml
DomAyre: Yes it's the cause of the consistent failure we've seen for the last few days, see the original PR #274

PR #285, opened 2025-01-09
Allow setting of JWT Issuer port

DomAyre:
### Motivation

Since #266, the test JWT issuer doesn't use a fixed port, privacy-sandbox-dev assumes it does which breaks it's testing.
To fix this the options are to either:

- Allow setting the port of the JWT issuer to the port assumed by privacy-sandbox-dev
- Read the full address in privacy-sandbox-dev

Since privacy-sandbox-dev assumes the port in many places (including in an instance running in ACI) it's non-trivial to do the first change, so opting for the latter.

### Changes

- [x] Update test JWT Issuer implementation to use the value of JWT_ISSUER_PORT as the port if set (default to 0 if unset)
- [x] Pass that env variable into the JWT issuer container image in docker compose
- [x] Set that value to 3000 for the standalone image used by privacy-sandbox-dev

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -50,4 +50,38 @@ WORKDIR /kms
 ARG CCF_PLATFORM
 ENV CCF_PLATFORM=${CCF_PLATFORM}
 ENV JWT_ISSUER_PORT=3000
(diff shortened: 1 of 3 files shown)


DomAyre:
privacy-sandbox-dev tests
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/12689008959

PR #287, opened 2025-01-09
Beejones/service request must show body in debug

beejones:
The body was not in the log during debug logging. Fixed


DomAyre:
Privacy-sandbox-dev fix is still in flight so I'll let you know when it's ready.

This PR looks like it's more about adding tests for the unwrap endpoint, can it be two separate PRs?

DomAyre:
It looks like there's still a lot of changes from #288 

I think you can drop changes in all files apart from:
- `ServiceRequest.ts`
- `ServiceRequest.test.ts`

src/utils/ServiceRequest.ts
```diff
@@ -95,18 +95,16 @@ export class ServiceRequest<T> {
...
-
     try {
       this.body = request.body.json();
     } catch (exception) {
-      this.error = {
-        errorMessage: `${this.logContext.getBaseScope()}: No valid JSON request for ${this.logContext.getFormattedScopeString()}`,
-      };
-      this.success = false;
-      return;
+      Logger.info("No JSON body found", this.logContext);
```
DomAyre, line 101: Looks like we're changing the behaviour from having no body being a critical failure to not caring, can you explain (maybe in a comment) why it's okay not to care?

beejones: The endpoints validate the properties they expect in the body. If the request did not pass a valid JSON body we have an empty body object and the endpoints typically return 400

beejones:
> Privacy-sandbox-dev fix is still in flight so I'll let you know when it's ready.
> 
> This PR looks like it's more about adding tests for the unwrap endpoint, can it be two separate PRs?

This is not related to only unwrapping (also /key ). While debugging unwrapping, I noticed the body was not listed in the Request debug message. This is now fixed.

beejones:
I was working on one machine on both PRs. They must have spoiled. I will fix it.

beejones:
> It looks like there's still a lot of changes from #288
> 
> I think you can drop changes in all files apart from:
> 
> * `ServiceRequest.ts`
> * `ServiceRequest.test.ts`

fixed

DomAyre (approved):

PR #288, opened 2025-01-10
Beejones/add key unwrapping to system tests

beejones:
Add unwrapKey test including decryption of the wrapped key


test/system-test/utils.py
```diff
@@ -173,10 +177,42 @@ def vote(proposal_id, vote, get_logs=False):
...
+def get_test_public_wrapping_key():
+    with open(os.path.join(os.path.dirname(__file__), '../data-samples/publicWrapKey.pem'), 'r') as file:
+        key = file.read()
+        # Escape newlines and wrap the key in double quotes
+        key = key.replace("\n", "\\n")
+        #return f"{key}"
+        return f'"{key}"'
+
+def get_test_private_wrapping_key():
+    with open(os.path.join(os.path.dirname(__file__), '../data-samples/privateWrapKey.pem'), 'r') as file:
```
DomAyre, line 189: Rather than two separate files for private and public key, could we have a single private key file and then for the get public key function we can derive the public key from the private key?

This would prevent the possibility of accidentally desyncing the two

test/system-test/utils.py
```diff
@@ -173,10 +177,42 @@ def vote(proposal_id, vote, get_logs=False):
...
+        key = file.read().strip()
+        return key
+"""
```
DomAyre, line 192: Why is this multi-line string here?

beejones: My original attempt to decrypt. It is still in comments. I will remove it later.

test/system-test/test_unwrapkey.py
```diff
@@ -0,0 +1,105 @@
...
+    assert status_code == 200
+
+    # unwrap key
+    status_code, unwrapped_json = unwrapKey(
+        attestation=get_test_attestation(),
+        wrapping_key=get_test_public_wrapping_key(),
+        wrappedKid=key_json["wrappedKid"]
+    )
+    assert status_code == 200
+    unwrapped = decrypted_wrapped_key(unwrapped_json["wrapped"])
```
DomAyre, line 28: It took me a second to remind myself of the unusual scheme of `/key` just returns a KID and then `/unwrapKey` returns the actual key (which by the looks of it is still wrapped?) 

Can you add a comment explaining what this test is expecting from the endpoint and how we're validating that it works correctly?

test/system-test/test_unwrapkey.py
```diff
@@ -0,0 +1,105 @@
...
+    assert status_code == 400
+
+
+def test_unwrap_key_missing_wrappedKid(setup_kms):
+    apply_kms_constitution()
+    apply_key_release_policy()
+    apply_settings_policy()
+    refresh()
+    while True:
+        status_code, key_json = key(
```
DomAyre, line 86: Why are we bothering to call `/key` when the test doesn't use it?

test/system-test/test_unwrapkey.py
DomAyre, line 1: Nice set of tests, can we also have one which tests what happens when you call the endpoint when there is no key in the KMS (aka before calling `/refresh`)?

test/system-test/utils.py
```diff
@@ -173,10 +177,42 @@ def vote(proposal_id, vote, get_logs=False):
...
-    return '"-----BEGIN PUBLIC KEY-----\\nMIICIjANBgkqhkiG9w0BAQEFAAOCAg8AMIICCgKCAgEA0L9FDBjydkdstv7OKqkw\\ndMiugRqlSHC9Lchfd7jh5uCzv602LhlBJQeEFYchvaEquISLQFoZxEkpGEbEb15v\\nN2dKwTCi0ioEGBidtFmuKiVZqf46Hbnw4OdinQHrlGO2PRsRE+DYPOy6xrZTKEnD\\nK+OnHDsZ0U2qNJ80IjbxcC83lQpaMx8Ij8AddGY9Msv0TMMgaVsrDaQYLC8tmJih\\nxVI2f5BHFbTy3pw4Sq9xp6se+S/ycOUF0M6RFArZcD9uR/NxJKFJJlGOKskgwLnT\\nlhPck4sGeLhHUydRdqw0+h8EDsTIUysNxMOYtZETaVuBS+ISMy8+WQgVPs12Ujr3\\n17kaHeZr8Lq0bwaFHruRBpNwqtUCBv57IBpe7hnEDDdvOvN/tPubf1dv3HxEt42T\\nqqfozS+a9+1hcI8hpNlEjh+qcy1BmhSOXvmRzlhauX4xv6OLCNkRxo6x2Q/1moDC\\nPgnaJJVIuESr07xnC8fk43i5qFzEXQO3hwNsjd7sqFBBTb6t5N6Nm37mMNTqsDky\\nniAeFG+1gK/UD+cMPfbUIDaCqpCDwrTX0gMqqUTDG6eNPmQaOa+slici3h9WLaNy\\nmEzfKqJMBggKib/+e4Eb/ENdvxeT1X2YXpZ3tjZE+bRoiDgN4FYqBzYtZ/ieRcsq\\n4fPqgZPbh+ivT2o7QutzWH0CAwEAAQ==\\n-----END PUBLIC KEY-----\\n"'  # pragma: allowlist secret
-
+def get_test_public_wrapping_key():
+    with open(os.path.join(os.path.dirname(__file__), '../data-samples/publicWrapKey.pem'), 'r') as file:
+        key = file.read()
+        # Escape newlines and wrap the key in double quotes
+        key = key.replace("\n", "\\n")
+        #return f"{key}"
```
DomAyre, line 185: I assume this is a typo?

test/system-test/utils.py
DomAyre: It's not a comment, it's a multiline string literal

test/system-test/utils.py
beejones: Is possible but not so obvious for PEM. With JWT it is very easy to do. 
We also use these files in bash scripts which would overcomplicate these.

DomAyre: In bash:
```
openssl rsa -in test/data-samples/privateWrapKey.pem -pubout
```

In python
```
RSA.import_key(get_test_private_wrapping_key()).publickey().export_key("PEM").decode()
```

test/system-test/test_unwrapkey.py
```diff
@@ -0,0 +1,105 @@
...
+    status_code, unwrapped_json = unwrapKey(
+        attestation=get_test_attestation(),
+        wrapping_key=get_test_public_wrapping_key(),
+        wrappedKid=key_json["wrappedKid"]
+    )
+    assert status_code == 200
+    unwrapped = decrypted_wrapped_key(unwrapped_json["wrapped"])
+    unwrapped_json = json.loads(unwrapped)
+    print(unwrapped_json)
+    assert unwrapped_json["kty"] == "OKP"
```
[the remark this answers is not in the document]
beejones: I don't believe so. If the decryption fails and not error out, it would return random data. We could not json decode that

test/system-test/test_unwrapkey.py
DomAyre: Done in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/288/commits/2809bbb0936ca82c4fa74cca193e3095bf7ea6dd

test/system-test/test_unwrapkey.py
DomAyre: Done in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/288/commits/2809bbb0936ca82c4fa74cca193e3095bf7ea6dd

test/system-test/test_unwrapkey.py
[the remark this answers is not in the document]
DomAyre: Fair enough, maybe we could enforce that it has all the fields that we expect? This code makes it look like we care more about the specific type rather than the fact we get a valid JWK. But up to you if you don't think it's worth the effort feel free to just resolve this one

test/system-test/test_unwrapkey.py
DomAyre: Addressed in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/288/commits/2809bbb0936ca82c4fa74cca193e3095bf7ea6dd

test/system-test/utils.py
beejones: it is a nice to have. We have several safety measures to detect a mismatch between private and public key. These new added tests will fail and also tests in privacy-sandbox-dev will fail.
My current priority is to unblock Muhamed. Let's move forward with the current implementation.

DomAyre: Sure

DomAyre (approved):

Issue #278, opened 2025-01-06
Use Ubuntu 24.04 for CI


DomAyre:
The field `runs-on` for the actions job which tests the KMS recently pointed at a new version of ubuntu (24.04)

This caused failures in all tests which involved azure, so #274 rolls the version back to the one that works.

However, since we will likely be chased onto 24.04 sooner or later we should fix the issues that stopped it working, a quick look shows that at least there is some issue with tokens expiring, the `install dependencies` step of the run takes 10m on 24.04 vs 1m on 22.04 so that is the likely culprit. We should probably look at what is installed there and reduce it to the essentials to speed up install time

PR #293, opened 2025-01-14
Use ubuntu 24.04 (latest) for system-test CI

takuro-sato:
To handle issue https://github.com/microsoft/azure-privacy-sandbox-kms/issues/278.
The problem reproduces in local containers too not only in github actions runner. ubuntu 24.04 struggles to run `pip install` even locally (it took around 10 minutes while it's 1 min with 20.04). It seems to be panda library which is especially slow.

The error in the CI is just token expiration, so I moved the login step just before it's required.


DomAyre:
This looks fine enough, I think it might also be a good idea to remove unnecessary dependencies

beejones (approved):

PR #294, opened 2025-01-15
Remove pandas and matplotlib

takuro-sato:
Looks like they are not used. We check it with the CI.


beejones:
we shpuld add matplotlib  to privacy sandbox dev so it can be removed from kms. Are you planning to do this?

DomAyre:
Superseded by #410

PR #297, opened 2025-01-17
dependabot updates

beejones:
Update dependabot alerts


beejones:
testing:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/12830863628

DomAyre:
Why is there a change in `e2e-test/src/index.ts` which isn't just version number changes

DomAyre:
It also looks like this PR breaks `privacy-sandbox-dev`

DomAyre (approved):

PR #310, opened 2025-02-06
stop redirecting for read only endpoints

beejones:
The redirects reveal the URL to the KMS endpoints and should not leak to the clients.
We stop redirecting for read-only endpoints to prevent the redirection strategy from revealing the URL to the KMS endpoints..


beejones:
tests
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/13183262182

takuro-sato (approved):

Issue #320, opened 2025-02-13
Run System Tests on ACL based KMS


DomAyre:
Once we can run KMS instances on ACL, the next step is to integrate that code with the system tests so that they can run against an ACL based KMS

PR #323, opened 2025-02-14
Run KMS Tests on ACL

DomAyre:
Addresses #320 

### Motivation

This is needed in order to have testing coverage of ACL based KMS's

### Changes

- [x] Test running a single test in Github Actions for an ACL based KMS
- [x] Check ACL costs to check it's viable to run all tests 
- [x] Run all tests
- [x] Implement a mechanism to mark expected failures when running on ACL (For tests using governance)
- [x] Fixed `pubkey.sh` helper script, `app.json` specifies that no authentication is required yet we attempt to provide a member cert, which unnecessarily breaks on ACL

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,8 +74,12 @@ jobs:
   system-stateful:
     name: system / stateful (local_keys)
     secrets: inherit # pragma: allowlist secret
+    strategy:
+      fail-fast: false
+      matrix:
+        env: [az-cleanroom-aci, acl]
     uses: ./.github/workflows/system-test.yml
     with:
       test_path: test_all_seq
-      env: az-cleanroom-aci
+      env: ${{ matrix.env }}
       use_akv: false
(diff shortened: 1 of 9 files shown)


DomAyre:
There was an unanticipated issue with running all tests, the bing ads subscription (used because it worked with az cleanroom) ran out of quota for ACL instances, so using the main DEV AND TEST subscription which is likely to have a higher quota

beejones (approved):

KenGordon (approved):

DomAyre:
Split out several changes to make this change simple
#330 
#331 
#329

KenGordon (approved):

DomAyre:
For future reference, the unexpected behaviour was caused by a difference in behaviour of a JQ version. 

In `propose.sh` there is a check if the output of the call to make a proposal has an error which is expected to also fail if there isn't JSON to parse. However in GitHub actions, the version of JQ doesn't fail when the response is just some whitespace, meaning the proposal doesn't fail as expected

Issue #332, opened 2025-02-24
Define application endpoints for changing the settings policy


DomAyre:
### Motivation

In order to support running the KMS on Azure Confidential Ledger (ACL), we need be able to configure the KMS without CCF governance.

DomAyre:
I've been exploring the possibility of mirroring the governance interface of CCF. The current process for this is:
- Set a constitution which contains an "action" which is two functions, validation for proposals and applying of proposals
- A proposal is then submitted (the validation function in the constitution is applied to check its a valid proposal
- Based on a resolution function defined in the constitution (e.g. auto accept/majority vote from members) the proposal is accepted (maybe after voting), and the values in the proposal are given to the apply function in the constitution.

So for a non CCF governance approach, we can mirror the interface at different levels:
- Everything including emulating setting of constitution
- Hard code the action into application logic, but have a "propose" endpoint which takes the proposal JSON as is
- Define endpoints which will have the validation and applying logic baked in

The more we mirror the CCF governance, the more flexibility we maintain, and the less we need to change in the calling logic, but the more we have to implement in the application code. 

So far, my personal preference is the middle option which would allow most of the proposal logic to stay the same (with a slightly different endpoint) but doesn't require implementing constitution setting which I think is fine given our application logic is known ahead of time anyway.

KenGordon:
I think I might need pictures to understand the options properly, any chance you can do some for the call on thursday? Nothing too detailed or fancy.

PR #335, opened 2025-02-25
Add proposals endpoint

DomAyre:
Partly Addresses #332 

### Why

In order to support ACL, operations which were previously done in the governance layer of CCF will now be done using application endpoints. This means we need an application endpoint which does the equivalent of CCFs `/gov/proposals`

### How
- [x] Implement a new application endpoint called `/proposals` which matches the `gov/proposals` ccf endpoint it will replace
- [x] Move the default CCF constitution to the KMS application (because it defines functions that our governance actions use)
  - NOTE: This is responsible for +1481 out of +1638 lines of this PR, but is identical except for some added `export` keywords where needed
- [x] Generate equivalent response JSON to the CCF governance

The next step is to use this endpoint to perform governance type operations


DomAyre:
I've added proposal_id and proposer_id. For now, these values are using the CCF digest option on the proposal and on the `request.caller` field

PR #337, opened 2025-02-26
Ensure down in conftest.py unsets env variables

DomAyre:
### Why

Sometimes `daily.yml` fails because we attempt to start new ACL instances while the old one is still "in" the system. This is because we re-use the deployment name from the last run.

### How

- [x] In python, remove the unset env variables in down.sh

### DIFF
--- a/test/system-test/conftest.py
+++ b/test/system-test/conftest.py
@@ -120,14 +120,8 @@ def _setup_ccf():
         },
     )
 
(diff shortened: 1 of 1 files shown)


DomAyre:
Here's a test run of `daily.yml`
https://github.com/microsoft/azure-privacy-sandbox-kms/actions/runs/13541839291

PR #340, opened 2025-02-26
Remove pandas dependency in requirements.txt

DomAyre:
### Why

We don't use it and installing it takes 10 minutes on all of our tests

- Unit tests go from 9m to 40s
- e2e tests go from 27m to 10m


KenGordon (commented):
It is always nice to see the back of pandas.

PR #343, opened 2025-02-27
Add tests for the `/auth` endpoint

DomAyre:
### Why

This is useful in particular for testing the change for moving JWT issuer trust from governance to application code

### How

- [x] Add script to call `/auth`
- [x] Add system tests which use the script
- [x] Clean up output directory for JWT issuer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ KMS_URL ?= https://127.0.0.1:8000
 KEYS_DIR ?= ${KMS_WORKSPACE}/sandbox_common
 RUN_BACK ?= true
 CCF_PLATFORM ?= virtual
-JWT_ISSUER_WORKSPACE ?= ${PWD}/jwt_issuer_workspace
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -6,7 +6,7 @@ KMS_URL ?= https://127.0.0.1:8000
 KEYS_DIR ?= ${KMS_WORKSPACE}/sandbox_common
 RUN_BACK ?= true
 CCF_PLATFORM ?= virtual
-JWT_ISSUER_WORKSPACE ?= ${PWD}/jwt_issuer_workspace
+JWT_ISSUER_WORKSPACE ?= ${PWD}/jwt_issuers_workspace/default
```
KenGordon, line 9: Why?

KenGordon (approved):

Makefile
DomAyre: It's point 3 on the "How" section, it just cleans up where the output of the JWT issuer goes because I noticed the makefile overrides the default in the script in a way where the files aren't properly ignored by git

PR #358, opened 2025-03-04
Move set_jwt_validation_policy into application code

DomAyre:
### Why

This is to support running the KMS on ACL which doesn't allow custom governance. Therefore anything which was previously custom governance must be moved to application code

### How

- [x] Move the proposal code from the KMS constitution actions file to the application code
- [x] In `jwt_issuer_trust.sh` call the proposals endpoint instead of submitting a governance proposal
- [x] Update the name of the map to not use the `ccf.gov` namespace
- [x] Fix the makefile to set the JS application code before attempting to trust the JWT issuer


KenGordon (approved):

src/actions/set_jwt_validation_policy.js
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+import { Action, checkType } from "./default_ccf.js";
+import { ccf } from "@microsoft/ccf-app/global";
+import { validationPolicyMapName } from "../authorization/jwt/JwtValidationPolicyMap";
+
+export let action = [
```
beejones, line 8: Are we using these as governance for sandbox?

beejones (approved):

src/actions/set_jwt_validation_policy.js
DomAyre: For the sake of documentation, Ronny missed a change showing the removal of the action code from `kms.js` and we confirmed in an offline chat that this PR is fully self contained for it's goal and doesn't rely on future PRs

PR #364, opened 2025-03-05
Use a single JWT validator

DomAyre:
### Why

The KMS today has a map from two hardcoded issuers:
- `http://Demo-jwt-issuer` which is our test JWT issuer
- `https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/` which is the issuer Azure AD v1 token endpoint

to separate implementations for handling JWTs, although the demo issuer code half reuses and half duplicates the other issuer code.

Not only is this redundant code and complexity, it means that if you attempt to propose `set_jwt_validation_policy` with a different issuer (e.g. the AD v2 endpoint) the proposal is applied, but then any attempt to use an endpoint with a token from the other issuer it will fail with 400 and no explanation.

I'm not sure if hardcoding the issuers we are allowed to propose trusting in the application code was intended, if it was we should catch at proposal time (I do see there is commented out code to do this, so that could just be uncommented). 

### How 

You can follow the first 9 commits to see the steps I took, each one is a trivial refactor. 

There is one commit (73b2ddafad2ee5d3785f206f363c961a5f62cec4) which uses lodash to just check that the JWT payload provided is a superset of the policy instead of manual code. I can imagine we might have deliberately chosen to avoid external dependencies, if that is the case, I think we can just implement a version of `isMatch()` in our own code, but I still think isolating the logic to check is worth having.


kapilvgit:
The intent was always to support a JWT policy that can be controlled using governance and made transparent on the ledger instead of being hard-coded in the application. This only impact access control, not security which is controlled through key release anyway.

src/authorization/jwt/JwtValidator.ts
```diff
@@ -4,55 +4,69 @@
...
+    }
+    Logger.debug(
+      `Validate JWT policy for issuer ${issuer}: ${JSON.stringify(policy)}`, this.logContext
     );
 
-    if (!provider) {
-      const error = `Authorization: JWT validation provider is undefined (JwtValidator) for ${<JwtIdentityProviderEnum>jwtCaller.jwt.keyIssuer}`;
-      Logger.error(error, this.logContext);
+    // Check that the JWT payload is a matching superset of the policy
+    if (!isMatch(jwtCaller.jwt.payload, policy)) {
```
DomAyre, line 55: We need to be able to support array values for token properties, we should go back to previous

DomAyre: Done in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/364/commits/2a703a431ce4327305b887af265ee972f273938b

src/authorization/jwt/JwtValidator.ts
```diff
@@ -4,55 +4,88 @@
...
+        {
+          errorMessage: "The JWT has no valid iss",
+          errorType: "AuthenticationError",
+        },
         400,
         this.logContext
       );
     }
-    const isValidJwtToken = provider.isValidJwtToken(jwtCaller);
+
```
beejones, line 34: reminder: we need to check if issuer is in our accepted issuers table.

DomAyre: I don't believe this code is functionally different at all from the previous code, so if this code is missing something, the original code is also missing something

beejones (approved):

PR #365, opened 2025-03-05
Revert version bumps

DomAyre:
### Why 

Privacy-sandbox-dev CI has been failing to build KMS. The logs claim it can't find the required version of pre-commit and matplotlib.

Quick solution is revert those bumps, I will also follow up with tests for building in the same way as privacy-sandbox-dev, and maybe investigating why it fails there and not here

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ docker==7.1.0
 GitPython==3.1.44
 httpx==0.27.2
 jinja2==3.1.5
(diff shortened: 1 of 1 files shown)


DomAyre:
Here's a run of privacy-sandbox-dev CI to see if we fix the issue
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/13674406828

kapilvgit (approved):

PR #374, opened 2025-03-06
Install pandas for sandbox running

DomAyre:
### Why

In privacy sandbox dev we see complaints that we're missing pandas


DomAyre:
Here's the privacy-sandbox-dev CI run
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/13695641554

Issue #322, opened 2025-02-13
Integrate ACL authentication


DomAyre:
ACL presents a slightly different authentication model with AAD based and cert based users (having a role of either Reader, Contributor or Administrator)

Currently, we just define the current AAD user who deploys the ACL and use a token based on that identity to make requests. However we should also support adding other users and specifying which user to make requests with

DomAyre:
Extra details for this issue after discussing with Sid:

- Roles are not enforced in user defined endpoints, and so we will need to update endpoints (maybe just `proposals` since that's the only endpoint which doesn't rely on attestation based auth)
- We can choose to define custom roles which can protect specific application tables

PR #377, opened 2025-03-06
Support and default to user cert authentication

DomAyre:
Partly addresses #322 

### Why

To support ACL, the concept of members and therefore member certs will no longer apply. Therefore for endpoints which currently default to member cert authentication but support user cert authentication (`/key` and `/unwrapKey`) we should use user certs by default.

### How

- [x] Update the script to bring up the sandbox CCF to create an initial user
- [x] Update the script to bring up ACL instances and add a cert based user
- [x] Add the `user_cert` option for auth in key.sh and unwrapKey.sh and default to it
- [x] Add user cert option and test in auth.sh]
- [x] Await the creation of the user in the ccf-sandbox healthcheck


scripts/kms/endpoints/key.sh
```diff
@@ -48,6 +48,8 @@ key() {
     auth_arg=()
     if [[ "$auth" == "member_cert" ]]; then
         auth_arg=(--cert $KMS_MEMBER_CERT_PATH --key $KMS_MEMBER_PRIVK_PATH)
+    elif [[ "$auth" == "user_cert" ]]; then
+        auth_arg=(--cert $KMS_USER_CERT_PATH --key $KMS_USER_PRIVK_PATH)
     elif [[ "$auth" == "jwt" ]]; then
         auth_arg=(-H "Authorization: Bearer $(curl -X POST $JWT_ISSUER | jq -r '.access_token')")
```
beejones, line 54: might be good to show an error if the wrong auth value is passed. Will be hard to debug without an error

test/e2e-test/src/index.ts
```diff
@@ -499,18 +499,6 @@ class Demo {
       keyResponse.gt["x-ms-ver"] === "1",
     );
 
-    // JWT not allowed
-    [statusCode, keyResponse] = await Api.keyReleasePolicy(
-      this.demoProps,
-      member,
```
beejones, line 505: use a cert instead of deleting the test

scripts/kms/endpoints/key.sh
DomAyre: What do you mean by wrong? The `/key` endpoint allows all three types of authentication, but if the user provides a value that the endpoint rejects shouldn't it already give you a suitable error?

test/e2e-test/src/index.ts
DomAyre: My understanding of this test is that we're checking that if you call `/keyReleasePolicy` with JWT auth it should fail, but this property is no longer the case.

It isn't the case for user cert either so what would changing to a user cert achieve?

beejones (approved):

DomAyre:
privacy-sandbox-dev CI:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/13793383170

PR #378, opened 2025-03-06
Fix the standalone image used for testing privacy-sandbox

DomAyre:
### Why

The existing standalone image for KMS (which is a single docker image which can run CCF and puts the KMS on top) is very complex and is currently failing.


DomAyre:
Privacy-sandbox-dev CI

https://github.com/microsoft/privacy-sandbox-dev/actions/runs/13701828596

PR #384, opened 2025-03-08
Fix privacy sandbox CI

DomAyre:
### Why

It was proving difficult to fix the privacy sandbox CI with a sweeping change, so reverted back with a minimal fix


DomAyre:
Privacy sandbox CI run:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/13740622204

PR #400, opened 2025-03-11
Don't build the KMS in the runtime image

DomAyre:
### Why

We build the KMS in a previous step and copy it over


DomAyre:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/13789851891

PR #404, opened 2025-03-11
Refactor running JWT issuer into function

DomAyre:


DomAyre:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/13792114755

PR #405, opened 2025-03-11
Merge `main` into `feature/app-based-gov`

DomAyre:
### Why

Main has had some changes which will be useful for app-based gov


DomAyre:
Fixing merge conflicts in #406

PR #408, opened 2025-03-12
Simplify the standalone KMS image

DomAyre:
### Why

We have more requirements than we need in requirements.txt just to satisfy the standalone image

PR #410, opened 2025-03-12
Remove unused dependencies

DomAyre:
### Why

With #408 we no longer need to have requirements for the sake of the standalone image, so we can now reduce the number of dependencies in requirements.txt, which in turn will reduce the noise from dependabot

### How

- [x] Remove implicit installation of requirements.txt when setting venv
- [x] Remove unused dependencies in requirements.txt 
- [x] Drop devcontainer building for e2e tests (drops e2e test time from 10m to 2m)


DomAyre:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/13815886995

DomAyre:
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/13853024313

KenGordon (approved):
Looks good. Nice PR description too :)

PR #412, opened 2025-03-14
Cleanup unused code

DomAyre:
### Why

While attempting to get #248 to work, I made some changes which are generally applicable, rather than throwing them away, here they are

### How

- [x] Bump node version
- [x] Remove unused scripts and config files
- [x] Remove unused dependencies
- [x] Remove commonjs


package.json
```diff
@@ -9,69 +9,43 @@
...
     "js-crypto-rsa": "1.0.7",
-    "jsonwebtoken": "9.0.2",
-    "node-fetch": "3.3.2",
     "rollup": "4.35.0",
     "ts-jest": "29.2.6",
     "ts-node": "10.9.2",
-    "tslib": "2.7.0",
-    "typescript": "5.7.3",
-    "uglify-js": "3.19.3"
+    "tslib": "^2.8.1",
```
beejones, line 35: we shouldn't use /\

rollup.config.js
```diff
@@ -13,10 +12,8 @@ export default {
...
   },
-  plugins: [nodeResolve(), typescript(), commonjs()],
-  onwarn: function (warning, rollupWarn) {
-    if (warning.code !== "THIS_IS_UNDEFINED") {
-      rollupWarn(warning);
-    }
-  },
+  plugins: [
+    nodeResolve(),
+    typescript()
```
beejones, line 17: why did we remove the warning and commonjs?

tsconfig.json
```diff
@@ -4,15 +4,14 @@
...
+    "noImplicitAny": false,
     "preserveConstEnums": true,
-    "sourceMap": false,
-    "resolveJsonModule": true,
-    "types": ["@types/node"],
-    "strict": true,
-    "noUnusedLocals": true
   },
-  "include": ["src/**/*", "test/**/*"]
+  "include": ["src/**/*"]
```
beejones, line 16: if I remember well the test part is needed otherwise the TS tests do not get build.

package.json
DomAyre: Thanks for that spot, I think it's a relic of the fact I removed it then re-added, so I'll restore it to the original state

rollup.config.js
DomAyre: For the warning, I don't think it's valuable to suppress undefined warnings.

For commonJS, that plugin is converting commonJS to ES modules, but all of our modules are already ES so it is extra complexity and potentially a source for confusion and errors

tsconfig.json
DomAyre: We run our tests with `ts-jest` so they don't need to be compiled with the src, you will see in the CI for this PR our tests run fine

package.json
DomAyre: I actually noticed it works with a slightly newer version so just dropping the caret

https://github.com/microsoft/azure-privacy-sandbox-kms/pull/412/commits/cc8ed83723fd4762c816e3718b251877b3581350

beejones (approved):

Issue #421, opened 2025-03-22
Allow setting JWT validation policy without needing local certs


DomAyre:
This will be used in production, because we want to bring up a system which uses AAD tokens, requiring we set the JWT validation policy, but we don't want to use local certs to do so and this is only suitable for testing

PR #425, opened 2025-03-24
Allow trusting JWT issuer without needing local certs

DomAyre:
Addresses #421 

### Why

There is a chicken and egg problem setting the JWT validation policy using a JWT, and since production KMS cannot use local certs, we need to have AKV based certs for this proposal.

Since we never want to store the keys locally, that means we need to support COSE signing the proposals much like we did for the `/gov/proposals` endpoint.

### How 
- [x] In `ccf/acl/up.sh`, switch on the value of `USE_AKV` and when set to true create identities in AKV
- [x] Update the KMS application code to be capable of handling `user_cose_sign1` authn_policies
  - [x] Create a new validator for this policy
  - [x] Parse out the body of the COSE sign1 document after validation
- [x] Default the proposals wrapper script to use COSE sign auth and update callers appropriately

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,7 @@ jobs:
         run: cat workspace/sandbox_0/out
 
   system-stateless:
-    name: system / stateless (${{ inputs.use_akv && 'akv' || 'local' }}_keys)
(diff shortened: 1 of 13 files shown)


DomAyre:
Despite the ACL tests being broken by removing programmability, this should be ready for review

DomAyre:
Replaced by #436

PR #431, opened 2025-03-25
Remove redundant settings policy setting

DomAyre:
### Why

While testing ACL based KMS's, Slava was led to believe by our tests for `/unwrapKey` that setting the settings policy is required for the endpoint to work, which it isn't. 

This lost him time and makes the tests expand their coverage outside of their intended domain unnecessarily, it would be confusing if these tests failed because setting the settings policy had a bug.

### How

Remove the calls to `set_settings_policy.sh`

### DIFF
--- a/test/system-test/test_unwrapkey.py
+++ b/test/system-test/test_unwrapkey.py
@@ -2,7 +2,7 @@ import json
 import os
 import pytest
 from endpoints import key, refresh, unwrapKey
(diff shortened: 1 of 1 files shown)


DomAyre:
Expect ACL tests to fail while programmability is broken

takuro-sato (approved):

PR #409, opened 2025-03-12
Merge feature/app-based-gov into main

DomAyre:
### Why

After branching `releases/0.1.0` from main, we can merge the branch `feature/app-based-gov` into main

PR #433, opened 2025-03-25
Remove constitution setting code

DomAyre:
### Why

As of #409, the KMS no longer has any custom actions in its CCF constitution. For production KMS will run on ACL which doesn't allow any customisation of the constitution.

Today, this code is either not called on ACL or always sets the default CCF constitution which it will already have, making it a no-op. 

### How
- [x] Remove all the code which sets the constitution
- [x] Remove all references to this code


Makefile
```diff
@@ -125,24 +121,12 @@ refresh-key: ## 🚀 Refresh a key on the instance
...
-	fi
-	@CCF_PLATFORM=${CCF_PLATFORM} ./scripts/submit_constitution.sh --network-url "${KMS_URL}" --certificate-dir "${KEYS_DIR}" --member-count ${MEMBER_COUNT}
-
 get-service-cert: # Get the mCCF service cert
 	@echo -e "\e[34m$@\e[0m" || true
 	$(call check_defined, IDENTITY_URL)
 	curl ${IDENTITY_URL} | jq ' .ledgerTlsCertificate' | xargs echo -e > ${KEYS_DIR}/service_cert.pem
 
-setup-mCCF: set-constitution deploy propose-add-key-release-policy propose-jwt-ms-validation-policy refresh-key  ## 🚀 Prepare an mCCF instance
+setup-mCCF: deploy propose-add-key-release-policy propose-jwt-ms-validation-policy refresh-key  ## 🚀 Prepare an mCCF instance
```
takuro-sato, line 129: Do we need this command at all now?

src/endpoints/proposals.ts
```diff
@@ -82,16 +82,27 @@ export const proposals = (
     if (isValidIdentity.failure) return isValidIdentity;
 
     // Handle ACL based authentication
```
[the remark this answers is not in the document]
DomAyre: @takuro-sato, this is somewhat out of sync with main so check #470

Makefile
DomAyre: Maybe, but I don't want to step on the toes of the other code in this PR more than is strictly necessary

DomAyre:
Moving to #470

PR #434, opened 2025-03-26
Default to JWT auth in testing

DomAyre:
### Why

When we support ACL with no local certs, this means many tests which rely on local certs will break. Those tests aren't specifically testing local certs so we can just default our authentication policy to JWT to avert this issue

### How
- [x] In the wrapper scripts which default to user or member cert auth, change that default to jwt
- [x] Update test setup to always trust the AAD JWT issuer
- [x] Remove a test which assumes no JWT issuer is currently trusted since that is now never the case

### DIFF
--- a/scripts/kms/endpoints/key.sh
+++ b/scripts/kms/endpoints/key.sh
@@ -5,7 +5,7 @@
 
 key() {
     params=()
-    auth="user_cert"
+    auth="jwt"
     attestation=""
     wrappingKey=""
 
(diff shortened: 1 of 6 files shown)


takuro-sato:
Does

takuro-sato:
Do we

takuro-sato:
Can we still run the test locally as before this PR?

DomAyre:
> Can we still run the test locally as before this PR?

I'm not sure which test you mean, but yes, we're just changing the default value for the auth parameter in the scripts which wraps calling the endpoints, but the endpoints themselves haven't changed at all, and you can still call those scripts with a different auth method

takuro-sato:
I just mean any test that is available locally. If the change requires extra steps (like setting env vars for AKV etc) to run tests, please update the doc.

takuro-sato (approved):

PR #435, opened 2025-03-26
Support AKV identities on ACL

DomAyre:
Partly addresses #421

### Why
In order to be able to COSE sign proposals with AKV identities, we first need the code to create AKV based identities. 

### How
- [x] Reuse the existing member-add code which can create AKV based identities
  - [x] This requires splitting member cert creation and adding to the network (this is because in ACL we need the identities before the creation of the network)


scripts/ccf/member/add.sh
```diff
@@ -5,43 +5,13 @@
...
 
-ccf-member-add() {
-    set -e
-
-    source $REPO_ROOT/scripts/ccf/az-cleanroom-aci/setup.sh && az-cleanroom-aci-setup
-    source $REPO_ROOT/scripts/ccf/propose.sh
-    source $REPO_ROOT/scripts/ccf/member/info.sh
-    source $REPO_ROOT/scripts/ccf/member/id.sh
-    source $REPO_ROOT/scripts/ccf/member/use.sh
+ccf-member-add-gov() {
```
takuro-sato, line 8: Can you add a brief comment on what it does? `-gov` part confused me at first, so a comment would help readers to understand it quickly

scripts/ccf/acl/up.sh
```diff
@@ -42,14 +28,10 @@ acl-up() {
...
     export KMS_MEMBER_PRIVK_PATH="$WORKSPACE/member0_privk.pem"
-    openssl ecparam -out "$KMS_MEMBER_PRIVK_PATH" -name "secp384r1" -genkey
-    openssl req -new -key "$KMS_MEMBER_PRIVK_PATH" -x509 -nodes -days 365 -out "$KMS_MEMBER_CERT_PATH" -"sha384" -subj=/CN="ACL Client Cert"
-
-    # Create a user cert
     export KMS_USER_CERT_PATH="$WORKSPACE/user0_cert.pem"
     export KMS_USER_PRIVK_PATH="$WORKSPACE/user0_privk.pem"
-    openssl ecparam -out "$KMS_USER_PRIVK_PATH" -name "secp384r1" -genkey
-    openssl req -new -key "$KMS_USER_PRIVK_PATH" -x509 -nodes -days 365 -out "$KMS_USER_CERT_PATH" -"sha384" -subj=/CN="ACL Client Cert"
+    ccf-member-create member0
```
takuro-sato, line 33: What happens if `acl-up` is called by accident (e.g. by carelessly invoking make commands)? Can we prevent overwriting the existing cert here?

scripts/ccf/member/create.sh
```diff
@@ -0,0 +1,36 @@
...
+    openssl req -new -key "$WORKSPACE/$1_privk.pem" -x509 -nodes -days 365 -out "$WORKSPACE/$1_cert.pem" -"sha384" -subj=/CN="KMS Client Cert"
+}
+
+ccf-member-create-akv() {
+    az keyvault certificate create \
+        --vault-name $AKV_VAULT_NAME \
+        --name $1 \
+        --policy "${AKV_POLICY:-$(cat $REPO_ROOT/scripts/akv/key_policy.json | jq -c .)}"
+    rm -rf $WORKSPACE/$1_cert.pem
+    az keyvault certificate download \
```
takuro-sato, line 20: If I remember correctly @beejones said that the keys in AKV never leave AKV as a requirement for the production. Can we keep handling the key like that?

test/system-test/utils.py
```diff
@@ -115,9 +115,16 @@ def trust_jwt_issuer(*args):
...
+    subprocess.run(
+        ["scripts/ccf/member/create.sh", member_name],
+        cwd=REPO_ROOT,
+        check=True,
+    )
+
 def add_member(member_name):
     subprocess.run(
-        ["scripts/ccf/member/add.sh", member_name],
+        ["scripts/ccf/member/add.sh", f"{os.getenv("WORKSPACE")}/{member_name}_cert.pem"],
```
takuro-sato, line 127: Probably docs/Adding_members.md should be updated?

scripts/ccf/member/add.sh
DomAyre: Sure will do

scripts/ccf/member/create.sh
DomAyre: This just downloads the certificate, the key stays in AKV

test/system-test/utils.py
DomAyre: Sure good spot

scripts/ccf/member/create.sh
takuro-sato: Ah, that's right.

scripts/ccf/acl/up.sh
DomAyre: Done in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/435/commits/3fa6b3ab8d36bd894a5791bb3dffe9767ed19037

scripts/ccf/member/add.sh
DomAyre: Added in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/435/commits/53644fddc2b27f8523ecdf812a70ed452a6b68b3

test/system-test/utils.py
DomAyre: https://github.com/microsoft/azure-privacy-sandbox-kms/pull/435/commits/139d70850c3ca620f8a428568b0a73eb314146bf

takuro-sato (approved):

Issue #430, opened 2025-03-25
Allow audit of proposal history


DomAyre:
Currently the new application endpoint `/proposals` doesn't save the proposals themselves, the thinking was since we have no voting mechanism we didn't need the table of active proposals.

However what we're missing is the auditability of *who* made the changes resulting from the proposal. So saving proposals will facilitate this.

We need to decide whether the proposal table is public or private.

If it is public:
- It's more transparent
- We would need to introduce defences against attackers taking those proposals and attempting to use them for replay and confusion attacks
  - The way CCF handles this is via two protected headers in the COSE document, `msg_type` and `created_at`, which we would need to parse out of the COSE document and enforce sensible policies against.

If it is private:
- We would need to think carefully about who should be able to audit it

DomAyre:
We had a meeting to discuss the options, the conclusion was to store proposals in a public table and therefore parse the protected headers in the same way CCF does

DomAyre:
I have been investigating the details of how this will work exactly, the goal of the design is to be as similar to the real CCF proposals as possible, unless it creates some headache for our use case.

I spoke with Eddy about how CCF does this, they have two governance tables, one for open proposals and one for "recent" COSE proposals. To audit the full history you read the ledger to see the history of the former table, and the latter table is used to enforce the check which defends against replay attacks. Eddy said the main reason they remove resolved proposals was mostly to deal with memory constraints on SGX.

My proposal for how we implement this is:
- We have a proposals table which stores the last n proposals (n being configurable maybe? likely at build time if so)
- We add a GET implementation to `/app/proposals` which returns the contents of the table
- For a full audit, users can inspect the raw ledger like they would in regular CCF

PR #436, opened 2025-03-26
Use COSE Sign1 authentication in proposals endpoint

DomAyre:
Partly addresses #421 and #430 

### Why

In order to be able to audit proposals history and make proposals with no local certs (before JWT validation policy is set), we need to be able to use COSE Sign1 authentication.

### How
- [x] Change `app.json` manifest to only accept cose authentication
- [x] Implement application code to check the COSE document
- [x] Update users of the proposals endpoint to COSE sign the proposal 
- [x] Fix up miscellaneous places which assumed certs would be local
- [x] Add testing for ACL and AKV together


beejones:
@DomAyre 
Is it possible to explain in a few lines how the trust model for the cose signature works for ACL and CCF?

DomAyre:
> @DomAyre Is it possible to explain in a few lines how the trust model for the cose signature works for ACL and CCF?

@beejones I've added a comment explaining where the validation code comes from, but I am hoping someone like you who knows more about the threat model and security nuances can check the logic (especially the assertion about CCF validating the document, I believe Amaury said this but I might have misunderstood)

Also worth considering our conversations in standup, which is also documented in #430 which would require further changes to this code in the future

DomAyre:
@beejones The note in the README was added in https://github.com/microsoft/azure-privacy-sandbox-kms/pull/436/commits/13d63a9902ac5ba2ba5b4a1581ffa0c1d8a7490f

Please let me know if you'd like to see any other changes

beejones (approved):

DomAyre:
@beejones Can you confirm you're happy with how the COSE document is validated? This is the main area I was hoping for review for on this PR

scripts/kms/key_rotation_policy_set.sh
```diff
@@ -14,9 +16,17 @@ key_rotation_policy_set() {
     # Echo the ROTATION_POLICY to the proposal file
     echo "$ROTATION_POLICY" | jq > "$WORKSPACE/proposals/set_key_rotation_policy.json"
 
-    # Submit the proposal
-    source $REPO_ROOT/scripts/kms/endpoints/proposals.sh \
-        "$WORKSPACE/proposals/set_key_rotation_policy.json"
+    (
+        # If running on sandbox_local, use the user cert because KMS can only
+        # authenticate user COSE signature
+        if [[ "$KMS_URL" == "https://127.0.0.1:8000" ]]; then
```
KenGordon, line 22: Nit:Since you have to do this more than once, would some flag indicating that the KMS was a special local one or some such scheme be nicer?

test/system-test/conftest.py
```diff
@@ -145,11 +145,21 @@ def setup_ccf_session():
 
 
 def _setup_kms():
-    if USE_AKV:
-        call_script([
-            "./scripts/akv/key-import.sh",
-            f'{os.getenv("DEPLOYMENT_NAME", "kms")}-private-key'
-        ])
+    if USE_AKV and TEST_ENVIRONMENT == "ccf/sandbox_local":
```
KenGordon, line 148: How does this relate to the check vs 127.0.0.1:8000? I guess this is a different scenario with a real AKV?

PR #438, opened 2025-03-28
Speed up start time for all ccf sandbox images

DomAyre:
### Why

Refactoring privacy-sandbox-dev means we can reuse the CCF image, this means we can make the performance optimisation to include SNP builds

### How

- [x] Derive python venv setup code from sandbox.sh and call it directly

### DIFF
--- a/services/ccf-sandbox/Dockerfile
+++ b/services/ccf-sandbox/Dockerfile
@@ -1,8 +1,16 @@
 ARG CCF_PLATFORM
-FROM ghcr.io/microsoft/ccf/app/dev/${CCF_PLATFORM}:ccf-5.0.11
+ARG CCF_VERSION=ccf-5.0.11
+FROM ghcr.io/microsoft/ccf/app/dev/${CCF_PLATFORM}:${CCF_VERSION}}
 
 # Pre install CCF dependencies to speed up container start (only works in virtual mode)
 ARG CCF_PLATFORM
 SHELL ["/bin/bash", "-c"]
 RUN apt-get install -y uuid-runtime gettext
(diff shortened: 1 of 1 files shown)


services/ccf-sandbox/Dockerfile
```diff
@@ -1,8 +1,16 @@
...
 ARG CCF_PLATFORM
 SHELL ["/bin/bash", "-c"]
 RUN apt-get install -y uuid-runtime gettext
-RUN (ls /opt/ccf_virtual/bin/sandbox.sh && (/opt/ccf_virtual/bin/sandbox.sh & until [[ $(curl -k -s -o /dev/null -w "%{http_code}" https://127.0.0.1:8000/node/network) =~ ^(200|400)$ ]]; do sleep 1; done)) || true
+
+# Prebuild python venv to speed up start time
+RUN python3 -m venv .venv_ccf_sandbox && \
+    . .venv_ccf_sandbox/bin/activate && \
+    pip install -U -q pip && \
+    VERSION=$(sed 's/^ccf-//' "/opt/ccf_${CCF_PLATFORM}/share/VERSION_LONG") && \
```
takuro-sato, line 14: Can we parse version from CCF_VERSION rather than using the a path that can be changed potentially?
I'm thinking defining CCF_VERSION like:
```
ARG CCF_VERSION=5.0.11
FROM ghcr.io/microsoft/ccf/app/dev/${CCF_PLATFORM}:ccf-${CCF_VERSION}
```

DomAyre: I considered this, but I decided using the file guarantees doing what the CCF implementation dictates not relying the current pattern holding

DomAyre: It could be any number of cases, maybe the python package and docker image versions desync, maybe the naming convention changes between the two completely. CCF sandbox explicitly checks this file as the source of truth for the version of the python package to use, so I think the safest thing for us to do is do what they do.

DomAyre: This piece of code in general is derived from `sandbox.sh`

PR #446, opened 2025-04-03
Refactor the standalone image to remove hardcoded proposals

DomAyre:
### Why

This is a step on the path to supporting ACL. 

Supporting ACL requires changing the dataplane shared libraries, in order to avoid rebuilding those images per test I want to merge https://github.com/KenGordon/data-plane-shared-libraries/pull/50. However, since main here and on privacy-sandbox-dev still relies on hard coding the proposal to trust our managed identity, that PR will not pass CI.

This PR alongside https://github.com/microsoft/privacy-sandbox-dev/pull/263 will fix that, allowing me to merge the dataplane PR and therefore iterate on ACL support more quickly

### How
- [x] Update the standalone image to use the newer scripts which don't hardcode any values

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,20 +22,11 @@ RUN npm install && npm run build
 # While KMS relies on running on sandbox.sh we need to use dev image
 # FROM mcr.microsoft.com/ccf/app/run-js/${CCF_PLATFORM}:${CCF_VERSION}
 FROM ghcr.io/microsoft/ccf/app/dev/${CCF_PLATFORM}:${CCF_VERSION}
-ARG CCF_PLATFORM
 WORKDIR /kms
 
 # Install nodejs for running the dummy JWT issuer
 RUN curl -fsSL https://deb.nodesource.com/setup_current.x | bash - && \
     apt-get install -y nodejs
-
-# Prebuild python venv to speed up start time
-RUN python3 -m venv .venv_ccf_sandbox && \
-    . .venv_ccf_sandbox/bin/activate && \
-    pip install -U -q pip && \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -64,20 +64,20 @@ start-idp:  ## 🏃 Start the idp for testing jwt
 # Start hosting the application using `sandbox.sh` and enable custom JWT authentication
 start-host: stop-host  ## 🏃 Start the CCF network using Sandbox.sh
 	@echo -e "\e[34m$@\e[0m" || true
-	MEMBER_COUNT=${MEMBER_COUNT} source ./scripts/ccf/sandbox_local/up.sh --build && \
-	source ./scripts/kms/js_app_set.sh && \
-	source ./scripts/kms/constitution_set.sh \
+	MEMBER_COUNT=${MEMBER_COUNT} . ./scripts/ccf/sandbox_local/up.sh --build && \
```
KenGordon, line 67: Have you go something against 'source" vs '.'? Or is there a particular difference I am missing?

Makefile
```diff
@@ -64,20 +64,20 @@ start-idp:  ## 🏃 Start the idp for testing jwt
...
+	./scripts/kms/js_app_set.sh && \
```
KenGordon, line 68: Did you mean to spawn a new shell? Is there any env variable being passed out? I guess this is deliberate but just in case...

KenGordon (approved):

Makefile
DomAyre: Nothing major, `.` is POSIX compliant whereas `source` is exclusive to bash and I think the code looks nicer personally

Makefile
DomAyre: There were no env variables being passes out by this so just calling the script directly is sufficient

PR #456, opened 2025-04-08
Support running privacy-sandbox-dev tests with ACL KMS

DomAyre:
### Why

This provides evidence that ACL based KMS's work with the B&A services.

### How

- [x] Allow using TEST_ENVIRONMENT to switch on ACL behaviour
  - This is needed because privacy-sandbox-dev wraps ACL in an NGINX proxy to use HTTP2
- [x] Update `.env` to not override existing env variables if already set
- [x] Fix check for existing certs 
- [x] Remove "standalone image" which was only used in privacy sandbox dev and is no longer needed


DomAyre:
Testing with accompanying privacy-sandbox-dev change
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/14352994559

DomAyre:
New test after merge
https://github.com/microsoft/privacy-sandbox-dev/actions/runs/14360777141

.github/workflows/ci.yml
```diff
@@ -68,13 +68,3 @@ jobs:
       test_path: test_all_seq
       env: acl
       use_akv: ${{ matrix.use_akv }}
-
-
-  standalone-image:
```
takuro-sato, line 73: Do we still have a CI that tests against local KMS after this PR?

DomAyre: Yes if you look at all the testing on this PR, all but 2 of the tests are with a local KMS, the removed test here was checking that the special image used by privacy-sandbox-dev (which runs CCF and sets up the KMS in one image instead of running a CCF sandbox container then configuring it later) builds

DomAyre: We remove that special image now because privacy-sandbox-dev now does what the rest of the KMS tests do, run CCF (either with a local container calling sandbox.sh, or ACL) and then sets up the KMS against it

takuro-sato (approved):

PR #457, opened 2025-04-11
Save accepted proposals

DomAyre:
Partly addresses #430 

### Why

Currently we don't have visibility of who called `/proposals` to make changes to KMS. This change saves the proposals into an application table

### How

- [x] Create a proposals table and save Proposals into the table
- [x] Add code for parsing out protected headers from COSE documents
- [x] Add check to `/proposals` that the current proposal was created after the last accepted proposal
- [x] Add a limit to how many proposals are kept

This is the scope for this PR, here is my suggestion for future PRs
- [x] Add testing for old proposals
  - [x] Make it easier to get historical proposals by adding a GET endpoint
- [ ] ~Defend against confusion attacks by checking `msg_type` field~ Not needed here because we don't have a voting system
- [ ] [Maybe] Make the number of proposal we keep configurable


package.json
```diff
@@ -16,6 +16,7 @@
     "@bufbuild/protobuf": "1.10.0",
     "@bufbuild/protoc-gen-es": "1.10.0",
     "@microsoft/ccf-app": "6.0.0",
+    "cbor-x": "1.6.0",
```
takuro-sato, line 19: Can we parse the header ourself or use a library made by a team close to us? While it's great there is such library, I think it's preferable if we can avoid external libraries to avoid situations like where we face a bug but we need to wait for fix long time etc.

src/endpoints/proposals.ts
```diff
@@ -135,7 +139,29 @@ export const proposals = (
...
+    // Create a map from created time to proposal ID for historical proposals
+    const createdTimeToProposalIdMap = new Map<number, ArrayBuffer>();
+    proposalsPolicyMap.forEach((proposal, proposalId) => {
+        createdTimeToProposalIdMap.set(
+            getCoseProtectedHeader(proposal)["ccf.gov.msg.created_at"],
+            proposalId
+        );
+    });
+
+    // Ensure the proposal was created after the last accepted proposal
```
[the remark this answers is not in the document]
DomAyre: I’ve got some documentation about this endpoint already at the top of the file, I’ve added some extra detail to answer this question: https://github.com/microsoft/azure-privacy-sandbox-kms/pull/457/commits/8027d05f3a9de21657a99642a19483372a637847

src/endpoints/proposals.ts
```diff
@@ -165,5 +191,19 @@ export const proposals = (
         ));
     }
 
+    // Save the proposal to the table
```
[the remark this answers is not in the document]
DomAyre: I’ve got some documentation about this endpoint already at the top of the file, I’ve added some extra detail to answer this question: https://github.com/microsoft/azure-privacy-sandbox-kms/pull/457/commits/8027d05f3a9de21657a99642a19483372a637847

package.json
DomAyre: If it was trivial to do either, I would also prefer that for the reasons you mentioned. Unfortunately I don’t think that is the case here. If I remember correctly from before I went away, @KenGordon was okay with us using a cbor library?

DomAyre:
> How is the change tested (to check correct contents are written etc)?

See the PR description for next PR steps, for now we have coverage that the existing functionality continues to work, then a future PR will add the ability to get old proposals which would then allow specific tests.

DomAyre:
@takuro-sato Don’t worry about the latest patch failing CI, it’s due to the issue fixed by #467

package.json
KenGordon: Obviously I would rather not but I don't think we want to be redoing this sort of thing ourselves either. It happens that these ought to be pretty simple so parsing a necessary subset would not be hard. However, plans like that often end up being fragile and so much more work in the end. On balance, I think this is ok.

takuro-sato:
> See the PR description for next PR steps, for now we have coverage that the existing functionality continues to work, then a future PR will add the ability to get old proposals which would then allow specific tests.

How about covering it within this PR? While I see the benefit of PRs being small, I think having test is better if there is a new feature. Test would help reviewers to understand code (e.g. to understand how it should behave for corner cases). Also generally we simply want to reduce the chance of main being broken (even if CI is green)

That's not a strong preference and I might easily be missing something

KenGordon (approved):

takuro-sato:
Having next change sounds good to me

test/system-test/test_proposals.py
```diff
@@ -0,0 +1,37 @@
+import json
+from endpoints import proposalsGet
+from utils import apply_key_release_policy
+from cose.messages import CoseMessage
+
```
takuro-sato, line 5: Can we have other test cases like:
- Check contents of proposalsGet() not only status code when there is no proposal
- Check behavior when more than 5 (`proposalsToKeep`) proposal are made. (check if it's actually old proposals that are deleted)
- Check the array is sorted as intended
?
-

src/endpoints/proposals.ts
```diff
@@ -165,5 +199,47 @@ export const proposals = (
...
     return ServiceResult.Succeeded<IProposalResult[]>(proposalResults, logContext);
+}
+
+export const getProposals = (
+    request: ccfapp.Request<IProposalsRequest>,
+  ): ServiceResult<string[]> => {
+    const logContext = new LogContext().appendScope("proposals");
+    let proposals: string[] = [];
+
+    // Sort the proposals by created time
```
takuro-sato, line 225: Is it because it's necessary? Asking just to understand

DomAyre: It isn't strictly necessary the client could do the sorting themselves, but I think it's more convenient to do it on the server side

test/system-test/test_proposals.py
DomAyre: The first case is hard to test because we will have a single proposal from setting up the KMS for testing.

The second one sounds good I'll add it

The third one is somewhat covered by the fact we pluck out proposals at specific indices and check which proposal it is, this would be super flaky if it wasn't sorted, checking the created at header isn't impossible but it would make the test infra more complex since we'd need to parse the cbor

src/endpoints/proposals.ts
takuro-sato: Got it thanks. You might want to document it somewhere "public" (vs implementation.) like app.json if you want to let clients take advantage of it (otherwise it's happened to be sorted from client point of view). But I think it's up to you.

DomAyre: Yeah if app.json had a "description" field that would be the perfect place, for now I've added it to the big comment at the top of proposals.ts

https://github.com/microsoft/azure-privacy-sandbox-kms/pull/457/commits/f348d89776afebb56ae53ad495d758203204e5c8

test/system-test/test_proposals.py
DomAyre: https://github.com/microsoft/azure-privacy-sandbox-kms/pull/457/commits/3b2095dfdca7040524e38d06e4c8a2d222d4407f

takuro-sato (approved):

